PR #2, opened 2020-08-18
Drtii 232 Improved login - upgrade & refactor

richbirch:
- Upgrade akka-http to 10.2.0
 - Upgrade akka-streams t0 2.6.8
 - Convert server to typed actor as per best practice examples
 - Refactor routes to separate classes


src/main/scala/uk/gov/homeoffice/drt/Server.scala
```diff
@@ -0,0 +1,73 @@
...
+    def starting(wasStopped: Boolean): Behaviors.Receive[Message] =
+      Behaviors.receiveMessage[Message] {
+        case StartFailed(cause) =>
+          throw new RuntimeException("Server failed to start", cause)
+        case Started(binding) =>
+          ctx.log.info(
+            "Server online at http://{}:{}/",
+            binding.localAddress.getHostString,
+            binding.localAddress.getPort)
+          if (wasStopped) ctx.self ! Stop
```
ngu04, line 63: Just understanding how it will work . 

will this gets executed ever ? as Stop is called when wasStopped = true and it is true when starting(wasStopped =true)

ngu04 (approved):
Overall looks good to me . Just question about one of things in actor below

src/main/scala/uk/gov/homeoffice/drt/Server.scala
richbirch: Yeah I've only just started getting my head around typed actors and changing their behaviours but it does make sense if you follow it through.
1) The actor is started with behaviour `starting(wasStopped = false)`
2) Before the server is started the actor is sent a `Stop` which is handled on line 65 and it changes its behaviour to `starting(wasStopped = true)`
3) The server starts and sends the actor a `Started` message which is handled in that code. Since `wasStopped` is now `true` that condition is executed and it sends itself a Stop and changes its behaviour to `running(binding)` so that the `Stop` message can handle the shutting down of the server

PR #4, opened 2020-11-26
Drtii 649 inaccessible ports for multi user setup

richbirch:
Deal with redirections from DRT when the user already has access to the port being requested


react/src/components/AccessRequestForm.tsx
```diff
@@ -68,9 +68,15 @@ export default class AccessRequestForm extends React.Component<IProps, IState> {
...
-    if (this.state.requestSubmitted)
-      content = <div>Thanks for your request. We'll get back to you shortly</div>;
-    else
+    if (this.state.requestSubmitted) {
+      setTimeout(function(){
+        window.location.href = "/oauth/logout?redirect=" + window.location;
+      }, 5000);
+      content = <div>
+        <p>Thanks for your request. We'll get back to you shortly></p>
+        <p>You will be logged out automatically in a moment</p>
```
bnppl: a lot of users might close the browser at this point, and then not get logged out, might be worth adding a link saying "click here to logout now" as well?

PR #8, opened 2021-01-28
Drtii 712 multiport alerts

bnppl:


react/src/App.tsx
```diff
@@ -1,47 +1,116 @@
...
+        this.state = {};
+    }
+
+    componentDidMount() {
+        this.apiClient.fetchData(this.apiClient.userEndPoint, this.updateUserState);
+        this.apiClient.fetchData(this.apiClient.configEndPoint, this.updateConfigState);
+    }
+
+    updateUserState = (response: AxiosResponse) => {
+        let user = response.data as UserLike;
```
richbirch: are we fussed about whether to use const wherever possible?

react/src/components/Alerts/SaveAlert.tsx
```diff
@@ -0,0 +1,172 @@
...
+                {}
+            ),
+            alertClass: "notice",
+            title: "",
+            message: "",
+            expires: moment().add(3, "hours").format("YYYY-MM-DDTHH:mm")
+        });
+
+    const updatePortSelection = (event: React.ChangeEvent<HTMLInputElement>) =>
+
```
richbirch: extraneous new line

react/src/components/Alerts/SaveAlert.tsx
```diff
@@ -0,0 +1,172 @@
...
+        setState({
+            ...state,
+            alertPorts: {
+                ...state.alertPorts,
+                [event.target.name]: event.target.checked
+            }
+        });
```
richbirch, line 71: missing new line

react/src/components/Alerts/SaveAlert.tsx
```diff
@@ -0,0 +1,172 @@
...
+    const updateField = (event: React.ChangeEvent<HTMLInputElement>) => setState({
```
richbirch: shall we put `setState` on its own line to match the others in this file?

react/src/components/Alerts/ViewAlerts.tsx
```diff
@@ -0,0 +1,100 @@
...
+        portAlerts: {} as IPortAlerts
+    });
+
+    const refreshAlerts = () => axios
+        .get("/api/alerts")
+        .then(response => setState({
+            portAlerts: response.data as IPortAlerts
+        }));
+
+    useEffect(() => {
```
richbirch, line 63: minor, but this could be on one line without the curlies, right?

react/src/services/ApiClient.ts
```diff
@@ -27,6 +28,6 @@ export default class ApiClient implements IApiClient {
 
   handleAjaxException(endPoint: string, throwable: any) {
     console.log('caught: ' + throwable);
-    window.document.location.reload();
+    // window.document.location.reload();
```
richbirch, line 31: I think I needed this reload for the case when we're not properly logged into keycloak. i can't remember exactly, but we should test everything works as expected for the sign up process if this line is removed

src/main/scala/uk/gov/homeoffice/drt/alerts/MultiPortAlert.scala
```diff
@@ -0,0 +1,36 @@
...
+
+case class MultiPortAlert(
+  title: String,
+  message: String,
+  alertClass: String,
+  expires: String,
+  alertPorts: Map[String, Boolean]) {
+
+  def alertForPorts(allPorts: List[String]): Map[String, Alert] = {
+    println(allPorts)
```
richbirch: printlns here

src/main/scala/uk/gov/homeoffice/drt/routes/ApiRoutes.scala
```diff
@@ -53,6 +71,84 @@ object ApiRoutes {
...
+        (post & path("alerts")) {
+          authByRole(CreateAlerts) {
+            import uk.gov.homeoffice.drt.alerts.MultiPortAlertJsonSupport._
+            headerValueByName("X-Auth-Roles") { rolesStr =>
+              headerValueByName("X-Auth-Email") { email =>
+                entity(as[MultiPortAlert]) {
+                  multiPortAlert =>
+                    {
+
+                      val futureResponses = multiPortAlert.alertForPorts(portCodes.toList).map {
```
richbirch: is it worth extracting this chunk to a function to keep the route code smaller and easier to read?

richbirch (changes requested):
Nice one. Just a handful of suggestions :+1:

react/src/services/ApiClient.ts
bnppl: it starts to cause an infinite reload loop pretty quickly if you have any connectivity issues, so may not be the best place for it, I forgot to figure out what to do with this though.

react/src/components/Alerts/ViewAlerts.tsx
bnppl: actually, no, `useEffect` requires a function with a void return type.

bnppl: (I tried it)

richbirch: :+1: good to know!

PR #22, opened 2021-06-25
Drtii 831 upload and process

ngu04:
- Nebo upload page
- Process file endpoint with akka stream
- Convert to json and post to DRT using akka-client


.gitignore
```diff
@@ -4,3 +4,4 @@ target/
 *.sc
 react/build
 src/main/resources/frontend
+.DS_Store
```
richbirch, line 7: It looks like the .DS_Store file already got committed. Could you remove it from git?

react/src/App.tsx
```diff
@@ -87,7 +87,7 @@ export default class App extends React.Component<IProps, IState> {
                     <Route exact path="/">
                         {this.state.config === undefined || this.state.user === undefined ?
                             <Loading/> :
-                            <Home config={this.state.config !!} user={this.state.user !! }/>
+                            <Home config={this.state.config !!} user={this.state.user !!}/>
```
richbirch: What are the `!!` for here? Googling tells me it's to convert an object into a boolean value, but that's when it's placed at the front of the object. It looks like they're maybe not doing anything

richbirch: If you drop the check for undefined then it'll also catch the null value whilst being a little shorter
```
{this.state.config && this.state.user ?
  <Home config={this.state.config} user={this.state.user}/> :
  <Loading/>
}
```

src/main/scala/uk/gov/homeoffice/drt/DrtClient.scala
```diff
@@ -0,0 +1,39 @@
...
+trait JsonSupport extends SprayJsonSupport with DefaultJsonProtocol {
+  implicit val flightDataFormat = jsonFormat4(FlightData)
+}
+
+trait HttpClient extends JsonSupport {
+  val log: Logger = LoggerFactory.getLogger(getClass)
+
+  def send(httpRequest: HttpRequest)(implicit executionContext: ExecutionContextExecutor, mat: Materializer): Future[HttpResponse]
+
+  def createDrtNeboRequest(data: List[FlightData], uri: String) = {
```
richbirch: can you add the return type definition for this function?

src/main/scala/uk/gov/homeoffice/drt/routes/IndexRoute.scala
```diff
@@ -26,10 +30,13 @@ case class IndexRoute(urls: Urls, indexResource: Route, directoryResource: Route
         staticResourceDirectory
       })
 
-  def indexRouteDirectives: Route = {
+  def indexRouteDirectives(path: String = ""): Route = {
```
richbirch: Can we change this string parameter to a neater implementation?
Maybe a new function that adds the redirection and common code factored out and shared? Or even just passing an object rather than a string. It feels better to use a new function as the behaviour is specific to this route

src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
```diff
@@ -0,0 +1,118 @@
...
+        complete("Not found!")
+      }
+      .result()
+
+  def apply(prefix: String, neboPortCodes: List[String], httpClient: HttpClient)(implicit ec: ExecutionContextExecutor, mat: Materializer): Route = {
+    val route: Route =
+      authByRole(NeboUpload) {
+        pathPrefix(prefix) {
+          post {
+            fileUpload("csv") {
```
richbirch: Might be worth putting this in a function to avoid the 10 levels of indentation and make it a little more readable

src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
```diff
@@ -0,0 +1,118 @@
...
+  def sendFlightDataToPort(flightData: Future[List[FlightData]], portCode: String, httpClient: HttpClient)(implicit ec: ExecutionContextExecutor, mat: Materializer): Future[(String, StatusCode)] = {
+    flightData.flatMap { fd =>
+      val httpRequest = httpClient
+        .createDrtNeboRequest(
+          fd.filter(_.portCode.toLowerCase == portCode.toLowerCase), s"${drtUriForPortCode(portCode)}$drtRoutePath")
+      httpClient.send(httpRequest)
+    }.map(r => portCode -> r.status)
+  }
+
+  def convertByteSourceToFlightData(metadata: FileInfo, byteSource: Source[ByteString, Any])(implicit ec: ExecutionContextExecutor, mat: Materializer) = {
```
richbirch: Add the return type

src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
```diff
@@ -0,0 +1,118 @@
...
+      indexMapRow.getOrElse(1, "").trim,
+      indexMapRow.getOrElse(2, "").trim,
+      indexMapRow.getOrElse(3, "").trim,
+      indexMapRow.getOrElse(4, "").trim)
+  }
+
+  private def rowToJson(rows: List[Row], metadata: FileInfo): List[FlightData] = {
+    val dataRows: Seq[Row] = rows.tail.filterNot(_.flightCode.isEmpty)
+    log.info(s"Processing ${dataRows.size} rows from the file name `${metadata.fileName}`")
+    dataRows.groupBy(_.arrivalPort)
```
richbirch, line 113: You can extract the tuples to readable variables to make it easier to see what's going on:

```
    dataRows.groupBy(_.arrivalPort)
      .flatMap { case (arrivalPort, portRows) =>
        portRows.groupBy(_.flightCode)
          .map { case (flightCode, flightRows) =>
            FlightData(arrivalPort, flightCode, covertDateTime(flightRows.head.date), flightRows.size)
          }
      }.toList```

File: src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala [hunk 8]
```diff
@@ -0,0 +1,118 @@
...
+      authByRole(NeboUpload) {
+        pathPrefix(prefix) {
+          post {
+            fileUpload("csv") {
+              case (metadata, byteSource) =>
+                onSuccess(
+                  Future.sequence(
+                    neboPortCodes
+                      .map(sendFlightDataToPort(
+                        convertByteSourceToFlightData(metadata, byteSource), _, httpClient))))(response => complete(s"Posted to DRT with status ${response.map(r => r._1 -> r._2)}"))
```
richbirch: Can we change the `complete` message to something more meaningful? I'm not sure what the status means and can't work it out from `r._1 -> r._2`
Also, what is `response` in this context?

src/test/scala/uk/gov/homeoffice/drt/routes/UploadRoutesSpec.scala
```diff
@@ -0,0 +1,86 @@
...
+  val multipartForm =
+    Multipart.FormData(Multipart.FormData.BodyPart.Strict(
+      "csv",
+      HttpEntity(ContentTypes.`text/plain(UTF-8)`, fileData),
+      Map("filename" -> "primes.csv")))
+
+  "Given a correct permission to users, the user should able to upload file successfully " >> {
+    Post("/uploadFile", multipartForm) ~>
+      RawHeader("X-Auth-Roles", NeboUpload.name) ~> RawHeader("X-Auth-Email", "my@email.com") ~> routes ~> check {
+        responseAs[String] shouldEqual """Posted to DRT with status List((lhr,202 Accepted))"""
```
richbirch: See previous comment about making `List((lhr,202 Accepted))` more meaningful

richbirch (changes requested):
Really good work mate. A few bits to tidy up, and some other suggestions, but very nice overall

react/src/App.tsx
ngu04: I think autoformat has done that . Change were not done by me here .

richbirch:
Oh yes! It's Mr Eppel to blame for this one :)

richbirch (approved):
wicked! :robot:

PR #23, opened 2021-06-28
Drtii 831 upload and process

ngu04:
- upload menu Item to be shown for user with nebo:upload permission


richbirch (approved):
thanks dude :rocket:

PR #24, opened 2021-06-29
Drtii 831 integration

ngu04:
- adding Port access role for Post DRT redlist passengers from nebo upload


src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
```diff
@@ -121,6 +122,28 @@ object UploadRoutes extends JsonSupport {
       }.toList
   }
 
+  private def portAccessRole(portCode: String): PortAccess = {
```
richbirch: We actually already have code to return a `Role` given a string. It's `Roles.parse` in drt-lib

richbirch (approved):
Nice one

PR #25, opened 2021-07-02
Drtii 831 departure date

ngu04:
- parsing new csv file with date and time separate and departure details
- fixing timezone issue with millis

### DIFF
--- a/src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
+++ b/src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
@@ -13,16 +13,16 @@ import org.slf4j.{ Logger, LoggerFactory }
 import spray.json._
 import uk.gov.homeoffice.drt.Dashboard._
 import uk.gov.homeoffice.drt.auth.Roles
-import uk.gov.homeoffice.drt.auth.Roles.{ NeboUpload, PortAccess, Role }
+import uk.gov.homeoffice.drt.auth.Roles.NeboUpload
 import uk.gov.homeoffice.drt.routes.ApiRoutes.authByRole
 import uk.gov.homeoffice.drt.routes.UploadRoutes.MillisSinceEpoch
 import uk.gov.homeoffice.drt.{ HttpClient, JsonSupport }
 
 import scala.concurrent.{ ExecutionContextExecutor, Future }
 
-case class Row(urnReference: String, associatedText: String, flightCode: String, arrivalPort: String, date: String)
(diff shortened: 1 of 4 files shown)


src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
```diff
@@ -97,34 +97,44 @@ object UploadRoutes extends JsonSupport {
...
+      arrivalTime = indexMapRow.getOrElse(5, "").trim,
+      departureDate = indexMapRow.getOrElse(6, "").trim,
+      departureTime = indexMapRow.getOrElse(7, "").trim,
+      embarkPort = indexMapRow.getOrElse(8, "").trim,
+      departurePort = indexMapRow.getOrElse(9, "").trim)
   }
 
   private def rowToJson(rows: List[Row], metadata: FileInfo): List[FlightData] = {
-    val dataRows: Seq[Row] = rows.tail.filterNot(_.flightCode.isEmpty)
+    val dataRows: Seq[Row] = rows.filterNot(_.flightCode.isEmpty).filterNot(_.flightCode == "Flight Code")
```
richbirch, line 117: It might be clearer that we're ignoring the header row if we use `drop(1)`, or reinstate using `tail` rather than filtering on the flight code

src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
```diff
@@ -97,34 +97,44 @@ object UploadRoutes extends JsonSupport {
...
             }
       }.toList
   }
 
-  val covertDateTime: String => MillisSinceEpoch = date => if (date.isEmpty) 0 else DateTime
-    .parse(date, DateTimeFormat.forPattern("dd/MM/yyyy HH:mm:ss"))
-    .withZone(DateTimeZone.UTC)
-    .getMillis
+  val covertDateTime: String => MillisSinceEpoch = date => if (date.isEmpty) 0 else
+    DateTimeFormat.forPattern("dd/MM/yyyy HH:mm").withZone(DateTimeZone.forID("Europe/London")).parseDateTime(date).getMillis
```
richbirch, line 138: :boom: nice!

richbirch (approved):
Nice - just one suggestion, but happy to merge :+1:

src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
ngu04: tail or drop is not going to work here as there some sequence happens due zipwithIndex and departureDate and schedule to have same value .

PR #28, opened 2021-07-22
Drtii 858

ngu04:
- updating error handling of nebo:file
- error for non Accepted status code
- error when no lines are parse
- multiple port support error display

### DIFF
--- a/react/src/components/FileUpload.tsx
+++ b/react/src/components/FileUpload.tsx
@@ -30,7 +30,7 @@ class FileUpload extends React.Component<IProps, IState> {
         this.state = {
             selectedFile: null,
             fileInput: React.createRef(),
-            displayMessage: [''],
+            displayMessage: [],
             hasError: false,
             errorMessage: '',
             showUploadButton: false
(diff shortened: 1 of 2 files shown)


react/src/components/FileUpload.tsx
```diff
@@ -49,40 +55,65 @@ class FileUpload extends React.Component<IProps, IState> {
...
     }
 
     responseData = (response: AxiosResponse) => {
-        console.log('response from post ' + response)
+        const feedStatusArray = response.data as FeedStatus[];
+        feedStatusArray.map(feedStatus => {
+            if (feedStatus.statusCode != '202 Accepted') {
+                this.setState({hasError: true});
+                this.setState({
+                    displayMessage: [...this.state.displayMessage, 'For port ' + feedStatus.portCode + ', ' + this.state.selectedFile.name + ' failed to upload. Check your file, try again later or contact us at drtpoiseteam@homeoffice.gov.uk.']
```
richbirch: Shall we move the message prefix to a function to remove the duplication? ie 'For port xxx, '

The email address is in config so we don't need to hard code it anywhere

richbirch: I'm thinking, if it's a non-202 status then that implies it was an issue with our app rather than the file, so we should probably just say to try again or contact us, rather than asking them to check their file

PR #29, opened 2021-07-27
Drtii 864 departure

ngu04:
- making departure details to be non mandatory while parsing csv file .

### DIFF
--- a/src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
+++ b/src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
@@ -129,8 +129,8 @@ object UploadRoutes extends JsonSupport {
                     FlightData(
                       portCode = arrivalPort,
                       flightCode = flightCode,
-                      scheduled = covertDateTime(arrivalDateTime),
-                      scheduledDeparture = flightRowsByArrival.head.departureDate.flatMap(dd => flightRowsByArrival.head.departureTime.map(dt => covertDateTime(s"$dd $dt"))),
+                      scheduled = parseDateToMillis(arrivalDateTime),
+                      scheduledDeparture = flightRowsByArrival.head.departureDate.flatMap(dd => flightRowsByArrival.head.departureTime.map(dt => parseDateToMillis(s"$dd $dt"))),
                       departurePort = flightRowsByArrival.head.departurePort.map(_.trim),
                       embarkPort = flightRowsByArrival.head.embarkPort.map(_.trim),
                       flightRowsByArrival.size)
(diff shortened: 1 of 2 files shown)


src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
```diff
@@ -107,12 +107,14 @@ object UploadRoutes extends JsonSupport {
...
-      departureTime = indexMapRow.getOrElse(7, "").trim,
-      embarkPort = indexMapRow.getOrElse(8, "").trim,
-      departurePort = indexMapRow.getOrElse(9, "").trim)
+      departureDate = emptyColumnCheck(indexMapRow.getOrElse(6, "")),
+      departureTime = emptyColumnCheck(indexMapRow.getOrElse(7, "")),
+      embarkPort = emptyColumnCheck(indexMapRow.getOrElse(8, "")),
+      departurePort = emptyColumnCheck(indexMapRow.getOrElse(9, "")))
   }
 
+  val emptyColumnCheck: String => Option[String] = column => if (column.isEmpty) None else Option(column)
```
richbirch: This function seems to be redundant if you change `getOrElse` to `get` in `indexMapRow.get(xxx)` would have the same result

src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
```diff
@@ -128,16 +130,16 @@ object UploadRoutes extends JsonSupport {
...
+                      departurePort = flightRowsByArrival.head.departurePort.map(_.trim),
+                      embarkPort = flightRowsByArrival.head.embarkPort.map(_.trim),
                       flightRowsByArrival.size)
                 }
             }
       }.toList
   }
 
-  val covertDateTime: String => MillisSinceEpoch = date => if (date.isEmpty) 0 else
+  val covertDateTime: String => MillisSinceEpoch = date =>
```
richbirch: i'd like this function to be be renamed to describe what it does. eg `parseDateToMillis`

richbirch (approved):
Good stuff. Just a couple of suggestions - a rename and probably function to remove to simplify the code :+1:

src/main/scala/uk/gov/homeoffice/drt/routes/UploadRoutes.scala
ngu04: One of file load failed as there was " " in the value of column . So this check is needed

richbirch: Ok. Let's rename it to something that says what it returns. eg `maybeColumnContent`

PR #33, opened 2021-09-17
Drtii 886 nebo all ports

ngu04:
- nebo file support for multiple ports

### DIFF
--- a/react/src/components/FileUpload.tsx
+++ b/react/src/components/FileUpload.tsx
@@ -12,7 +12,7 @@ interface IProps {
 interface IState {
     selectedFile: any;
     fileInput: any;
-    displayMessage: string[];
+    displayMessage: string;
     hasError: boolean;
     errorMessage: string;
     showUploadButton: boolean;
@@ -31,7 +31,7 @@ class FileUpload extends React.Component<IProps, IState> {
         this.state = {
             selectedFile: null,
             fileInput: React.createRef(),
-            displayMessage: [],
+            displayMessage: '',
             hasError: false,
             errorMessage: '',
             showUploadButton: false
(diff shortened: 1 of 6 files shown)


react/src/components/FileUpload.tsx
```diff
@@ -78,43 +78,48 @@ class FileUpload extends React.Component<IProps, IState> {
...
-   generateMessage(portCode:string ,fileName:String, message:string) {
-        return 'For port ' + portCode + ', ' + fileName + message ;
-   }
+    generateMessage(portCode: string, fileName: String, message: string) {
+        return 'For port ' + portCode + ', ' + fileName + message;
+    }
 
     responseData = (response: AxiosResponse) => {
         const feedStatusArray = response.data as FeedStatus[];
+        var failedPorts: string = ' ';
```
richbirch: we could eliminate some repeated code here by leveraging `filter` and `join`:
```
const failedStatuses = feedStatusArray.filter(s => s.statusCode != '202 Accepted' || s.flightCount == '0')
const failedPorts = failedStatuses.map(s => s.portCode).join(', ')
feedStatusArray.map(s => console.log('response feed ' + s.portCode + ' ' + s.flightCount + ' ' + s.statusCode))
if (failedStatuses.length > 0) this.setState({hasError: true})
```

react/src/components/FileUpload.tsx
```diff
@@ -78,43 +78,48 @@ class FileUpload extends React.Component<IProps, IState> {
...
-                        displayMessage: [...this.state.displayMessage, this.generateMessage(feedStatus.portCode,this.state.selectedFile.name,' Arrivals have been updated. Thank you!')]
-                    });
+                    console.log('response feed ' + feedStatus.portCode + ' ' + feedStatus.flightCount + ' ' + feedStatus.statusCode);
                 }
             }
             console.log('response feed ' + feedStatus.portCode + ' ' + feedStatus.flightCount + ' ' + feedStatus.statusCode);
         });
+
+        if (this.state.hasError) {
+            let m = 'The file has been uploaded, but there was a problem with ' + failedPorts.substring(0, failedPorts.length - 2) + ' ports. Please contact the DRT team for further information.';
```
richbirch: can we use `const` rather than `let` for good practice here?

richbirch (approved):
Good stuff. There's a bit of suggested tidying to make the code more succinct and reduce repeated lines of code

PR #34, opened 2021-09-17
Drtii 803 red list admin UI

richbirch:
Add red list editor
Implement end point to broadcast red list changes to all ports
Tidy up templates and state by introducing Redux
Improve dashboard layout


src/main/scala/uk/gov/homeoffice/drt/Server.scala
```diff
@@ -50,8 +51,9 @@ object Server {
         staticResourceDirectory = getFromResourceDirectory("frontend/static")).route,
       CiriumRoutes("cirium", serverConfig.ciriumDataUri),
       DrtRoutes("drt", serverConfig.portCodes),
-      ApiRoutes("api", serverConfig.portCodes, serverConfig.rootDomain, notifications, serverConfig.teamEmail),
-      UploadRoutes("uploadFile", serverConfig.neboPortCodes.toList, new DrtClient))
+      ApiRoutes("api", serverConfig.portCodes, serverConfig.rootDomain, notifications, serverConfig.teamEmail, neboRoutes),
+//      NeboUploadRoutes(serverConfig.neboPortCodes.toList, new ProdHttpClient)
```
ngu04: is it comment by mistake ?

src/main/scala/uk/gov/homeoffice/drt/routes/ApiRoutes.scala
```diff
@@ -95,34 +100,85 @@ object ApiRoutes {
...
           }
         },
+        (post & path("red-list-updates")) {
+          authByRole(RedListsEdit) {
+            import spray.json._
+            import uk.gov.homeoffice.drt.redlist.RedListJsonFormats._
+            entity(as[SetRedListUpdate]) {
+              setRedListUpdate =>
+                val portRoles = Roles.portRoles
+                Seq(LHR).map { portRole =>
```
ngu04: is it map to LHR only here ? do we not need all ports here ?

src/main/scala/uk/gov/homeoffice/drt/routes/ApiRoutes.scala
```diff
@@ -95,34 +100,85 @@ object ApiRoutes {
...
+                    setRedListUpdate.toJson.compactPrint,
+                    Seq(RedListsEdit, portRole))
+                }
+                complete(Future(StatusCodes.OK))
+            }
+          }
+        },
+        (get & path("red-list-updates")) {
+          authByRole(RedListsEdit) {
+            implicit val rlJson: RedListJsonFormats.redListUpdateJsonFormat.type = uk.gov.homeoffice.drt.redlist.RedListJsonFormats.redListUpdateJsonFormat
```
ngu04, line 121: It would nice to import the class uk.gov.homeoffice.drt.redlist.RedListJsonFormats and just use RedListJsonFormats. redListUpdateJsonFormat or redListUpdateJsonFormat

src/main/scala/uk/gov/homeoffice/drt/routes/ApiRoutes.scala
```diff
@@ -95,34 +100,85 @@ object ApiRoutes {
...
+            implicit val rlsJson: RedListJsonFormats.redListUpdatesJsonFormat.type = uk.gov.homeoffice.drt.redlist.RedListJsonFormats.redListUpdatesJsonFormat
+            val requestPortRole = LHR
```
ngu04, line 123: same here , it is just for  LHR .

src/test/scala/uk/gov/homeoffice/drt/routes/NeboUploadRoutesSpec.scala
```diff
@@ -89,65 +88,66 @@ class UploadRoutesSpec extends Specification with Specs2RouteTest {
...
       "csv",
       HttpEntity(ContentTypes.`text/plain(UTF-8)`, test1FileData),
       Map("filename" -> "test1.csv")))
 
   "Given a correct permission to users, the user should able to upload file successfully " >> {
-    Post("/uploadFile", multipartForm) ~>
-      RawHeader("X-Auth-Roles", NeboUpload.name) ~> RawHeader("X-Auth-Email", "my@email.com") ~> routes ~> check {
-        responseAs[String] shouldEqual """[{"flightCount":1,"portCode":"lhr","statusCode":"202 Accepted"}]"""
-      }
+    Post("/", multipartForm) ~>
```
ngu04: would not be good be keep /neboUpload rather than root ? just suggestion

react/src/components/RedListEditor.tsx
```diff
@@ -0,0 +1,340 @@
...
+  const saveEdit = (editing: Editing) => {
+    const withoutOriginal = state.editing && state.updates.filter(u => u.effectiveFrom !== state.editing?.update.effectiveFrom && u.effectiveFrom !== state.editing?.originalDate)
+    const withNew = withoutOriginal && state.editing && withoutOriginal.concat(state.editing.update)
+    console.log('Updated state with saved edit. TODO: call endpoints to persist to all ports')
+    const request: RequestSetRedListUpdates = {
+      updates: {
+        originalDate: editing.originalDate,
+        redListUpdate: editing.update
+      },
+      onSuccess: () => {
```
ngu04, line 105: May can have console log on success.

ngu04 (approved):
Nice one @richbirch ! Few minor comments

react/src/components/RedListEditor.tsx
richbirch: Yeah. I was actually planning to use a toaster to pop up a message for success and failure. I didn't mean to leave this unused whoops!

src/main/scala/uk/gov/homeoffice/drt/Server.scala
richbirch: Yes a mistake sorry! We need to refactor the routes as they're very fragile at the moment

src/main/scala/uk/gov/homeoffice/drt/routes/ApiRoutes.scala
richbirch: Yes this need replacing with the port roles val above. I had naughtily hard coded LHR for testing locally

src/main/scala/uk/gov/homeoffice/drt/routes/ApiRoutes.scala
richbirch: I would have just imported the formats class but for a reason I could not fathom it just wouldn't work that way, hence I had to resort to explicitly importing the implicits, which is ridiculous. I spent a long time trying to figure out why it wouldn't work the other way when all the other json format objects work fine...

src/main/scala/uk/gov/homeoffice/drt/routes/ApiRoutes.scala
richbirch: This one is fine using just LHR because we're reading the red lists in order to display them for editing in the dashboard app. It's not ideal having them stored for each port individually, but I didn't want to bite off that extra bit of work to store them centrally in this ticket

src/test/scala/uk/gov/homeoffice/drt/routes/NeboUploadRoutesSpec.scala
richbirch: This is due to refactoring the nebo route so that it doesn't contain the prefix, hence in the tests we don't need to request a specific url as the route will catch any url

src/main/scala/uk/gov/homeoffice/drt/Server.scala
richbirch: Ah sorry, this wasn't a mistake afterall. I've moved the nebo route to be used within the api routes, as it is an API. I do need to remove the comment though

PR #57, opened 2022-02-18
Drtii 924 group ports by region

richbirch:
Display port list by region on homepage when user has access to ports in multiple regions
Display port lists by region on signup page 
 - Allow easy selection of all ports a region
 - Allow easy selection of all ports in the UK
Clean up header layout
Use flexing layout for regions that wrap with smaller screens

### DIFF
--- a/src/main/scala/uk/gov/homeoffice/drt/json/PortRegionJsonFormats.scala
+++ b/src/main/scala/uk/gov/homeoffice/drt/json/PortRegionJsonFormats.scala
@@ -1,0 +1,14 @@
+package uk.gov.homeoffice.drt.json
+
+import spray.json.{ DefaultJsonProtocol, JsArray, JsObject, JsString, JsValue, RootJsonFormat, enrichAny }
+import uk.gov.homeoffice.drt.ports.PortRegion
+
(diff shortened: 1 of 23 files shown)


src/main/scala/uk/gov/homeoffice/drt/json/PortRegionJsonFormats.scala
```diff
@@ -0,0 +1,14 @@
+package uk.gov.homeoffice.drt.json
+
+import spray.json.{ DefaultJsonProtocol, JsArray, JsObject, JsString, JsValue, RootJsonFormat, enrichAny }
+import uk.gov.homeoffice.drt.ports.PortRegion
+
+object PortRegionJsonFormats extends DefaultJsonProtocol {
+  implicit object PortRegionJsonFormat extends RootJsonFormat[PortRegion] {
+    override def read(json: JsValue): PortRegion = throw new Exception("Deserialising PortRegion not implemented yet")
```
ngu04, line 8: Optional Comment  : May be should add read Deserialize . Could be useful for test further .

ngu04 (approved):
🏈

PR #69, opened 2022-08-09
Drtii 1002

ngu04:
- Regional Dashboard Arrival export


react/src/components/ExportDatePicker.tsx
```diff
@@ -0,0 +1,53 @@
...
+interface IProps {
+    region: string;
+}
+
+export default function ExportDatePicker(props: IProps) {
+    const [fromValue, setFromValue] = React.useState<Date | null>(null);
+    const [toValue, setToValue] = React.useState<Date | null>(null);
+    const formattedDate = (date: Date) => format(date as Date, "yyyy-MM-dd")
+
+    const determineDisplay = () => {
```
richbirch: `determineDisplay` could be renamed to be descriptive of what it does. Maybe something like `renderDownloadButton`?

react/src/components/ExportDatePicker.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <LocalizationProvider dateAdapter={AdapterDateFns}>
+            <DatePicker
+                label="From Date"
+                value={fromValue}
+                onChange={(newValue) => {
+                    setFromValue(newValue);
+                }}
+                renderInput={(params) => <TextField {...params} />}
+            />
+            <span>&nbsp;&nbsp;</span>
```
richbirch: Rather than using `&nbsp;` it'd be good to harness some css to get the spacing right. Flexbox is good for this as you can specify the gap between items in the container and have then align nicely too.

react/src/components/ExportDatePicker.tsx
```diff
@@ -0,0 +1,53 @@
...
+            <DatePicker
+                label="To Date"
+                value={toValue}
+                onChange={(newValue) => {
+                    setToValue(newValue);
+                }}
+                renderInput={(params) => <TextField {...params} />}
+            />
+            <br/>
```
richbirch: Same here - let's use some css to create the gap rather than `<br/>`s

react/src/components/PortListByRegion.tsx
```diff
@@ -8,40 +7,46 @@ import Icon from "@mui/icons-material/FlightLand";
...
-              </ListItem>
-              <Divider variant="inset" component="li"/>
+    >
+        {props.regions.map((region) => {
+            const sortedPorts = region.ports.sort()
+            return <Box sx={{
+                minWidth: '200px',
+                marginRight: '50px'
+            }}>
+                {props.user.roles.includes("rcc:view") ?
```
richbirch: We'll need region specific roles rather than a single rcc role, just like we have a role for each port. Most RCCU users will belong to a single RCCU so should only have access to their own region

react/src/components/RegionalPort.tsx
```diff
@@ -0,0 +1,40 @@
...
+                    <div align="left"><ArrivalExport region={props.region}/></div>
+                    <p> A member of the team will be in touch to get your thoughts about how this
+                        page could be more useful. You can also get in touch with the team by email at
+                        <a href="mailto:drtpoiseteam@homeoffice.gov.uk"> drtpoiseteam@homeoffice.gov.uk</a>.</p>
+                </Box>
+                <Button style={{float: 'right'}} href="/">back</Button>
+            </div> :
+            <div>
+                <Box>
+                    <p> You don't have relevant permission to view the page. If you like to have permission for the page
```
richbirch: `You don't have access to this page. To request access please get in touch with the team at drtpoiseteam@homeoffice.gov.uk.`

react/src/components/RegionalPort.tsx
```diff
@@ -0,0 +1,40 @@
...
+                <Box>
+                    <Typography align="left" variant="h6" component="h2">
+                        {props.region} region dashboard
+                    </Typography>
+                    <p> This is a new page on DRT. You can download an arrivals export covering all port terminals in
+                        this region.</p>
+                    <div align="left"><ArrivalExport region={props.region}/></div>
+                    <p> A member of the team will be in touch to get your thoughts about how this
+                        page could be more useful. You can also get in touch with the team by email at
+                        <a href="mailto:drtpoiseteam@homeoffice.gov.uk"> drtpoiseteam@homeoffice.gov.uk</a>.</p>
```
richbirch: Can we un-hardcode the email address? It might already be available in the config but if not we can add it

src/main/scala/uk/gov/homeoffice/drt/rccu/ExportCsvService.scala
```diff
@@ -0,0 +1,68 @@
...
+
+class ExportCsvService(httpClient: HttpClient) {
+
+  val log: Logger = LoggerFactory.getLogger(getClass)
+
+  def drtUriForPortCode(portCode: String): String = s"http://${portCode.toLowerCase}:9000"
+
+  val drtExportCsvRoutePath = "export/arrivals"
+
+  def getPortRegion(region: String): PortRegion = PortRegion.regions.filter(_.name == region).head
```
richbirch: `.head` will throw an exception if the region isn't found. You could use `headOption` and handle the case where the region isn't found

src/main/scala/uk/gov/homeoffice/drt/routes/ExportRoutes.scala
```diff
@@ -0,0 +1,57 @@
...
+    path("export" / Segment / Segment / Segment) { (region, startDate, endDate) =>
+      val portRegion: PortRegion = exportCsvService.getPortRegion(region)
+      val fileName = exportCsvService.makeFileName(startDate, endDate, region)
+      val portCsvHeadings = "IATA,ICAO,Origin,Gate/Stand,Status,Scheduled,Est Arrival,Act Arrival,Est Chox,Act Chox,Minutes off scheduled,Est PCP,Total Pax,PCP Pax,Invalid API,API e-Gates,API EEA,API Non-EEA,API Fast Track,Historical e-Gates,Historical EEA,Historical Non-EEA,Historical Fast Track,Terminal Average e-Gates,Terminal Average EEA,Terminal Average Non-EEA,Terminal Average Fast Track,API Actual - EEA Machine Readable to e-Gates,API Actual - EEA Machine Readable to EEA,API Actual - EEA Non-Machine Readable to EEA,API Actual - EEA Child to EEA,API Actual - GBR National to e-Gates,API Actual - GBR National to EEA,API Actual - GBR National Child to EEA,API Actual - B5J+ National to e-Gates,API Actual - B5J+ National to EEA,API Actual - B5J+ Child to EEA,API Actual - Visa National to Non-EEA,API Actual - Non-Visa National to Non-EEA,API Actual - Visa National to Fast Track,API Actual - Non-Visa National to Fast Track,Nationalities,Ages"
```
richbirch: Probably for a future ticket - we should probably specify the column configuration in a single place, drt-lib, so that it remains in sync when we come to change it in future and prevent this export breaking

src/test/scala/uk/gov/homeoffice/drt/rccu/ExportCsvServiceSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  "Get response for given region port terminal" >> {
+    val portResponse: Future[Option[PortResponse]] = exportCsvService.getPortResponseForTerminal("2022-07-22", "2022-07-23", "Heathrow", "Heathrow", "T2")
+    val response = Await.result(portResponse, 1 seconds)
+    response must beSome(PortResponse("Heathrow", "Heathrow", "T2", HttpResponse(entity = HttpEntity(ContentTypes.`text/csv(UTF-8)`, csv))))
+
+  }
+
+  object MockHttpClient extends HttpClient {
```
richbirch, line 40: We have 2 copies of this mock class. We can either make them specific to the tests or unify them to a single class that can be used for any test

src/test/scala/uk/gov/homeoffice/drt/routes/ExportRoutesSpec.scala
```diff
@@ -0,0 +1,159 @@
...
+      Get("/export/Central/2022-08-02/2022-08-03") ~> ExportRoutes(MockHttpClient) ~> check {
+        val a = responseAs[String]
+        a should ===(centralRegionPortTerminalData)
+      }
+    }
+  }
+
+  object MockHttpClient extends HttpClient {
+    def send(httpRequest: HttpRequest)(implicit executionContext: ExecutionContextExecutor, mat: Materializer): Future[HttpResponse] = {
+      if (httpRequest.headers.mkString.contains("pik")) {
```
richbirch: It doesn't look like this `pik` specific code is used in any of the tests

richbirch (changes requested):
Nice one mate. Just a few places to tidy things up :boom:

src/test/scala/uk/gov/homeoffice/drt/routes/ExportRoutesSpec.scala
ngu04: This is just to cover the scenario where port is not available .

react/src/components/PortList.tsx
```diff
@@ -3,27 +3,35 @@ import {Box} from "@mui/material";
...
+        return {...region, ports: userPorts} as PortRegion
+    }).filter(r => r.ports.length > 0)
 
-  return <Box sx={{width: '100%'}}>
-    <h1>Welcome to DRT</h1>
-    <p>Select your destination</p>
-    {userRegions.length > 1 ?
-      <PortListByRegion regions={userRegions} drtDomain={props.drtDomain}/> :
-      <PortListStraight ports={props.userPorts} drtDomain={props.drtDomain}/>
+    const checkRccRoles = () => {
```
richbirch: This name is a little ambiguous. Could be something like `isRccUser`

src/test/scala/uk/gov/homeoffice/drt/rccu/ExportCsvServiceSpec.scala
richbirch: This PIK code isn't used as far as I can see. How come you want to keep it?

richbirch (approved):
Nice one. Just minor comments

PR #70, opened 2022-08-15
Drtii 1061

ngu04:
- request for rccu region access

### DIFF
--- a/react/src/components/AccessRequestForm.tsx
+++ b/react/src/components/AccessRequestForm.tsx
@@ -10,151 +10,167 @@ import ListItemText from "@mui/material/ListItemText";
 import {Box, Button, Divider, FormControl, TextField, Typography} from "@mui/material";
 import {PortRegion, PortRegionHelper} from "../model/Config";
 import {PortsByRegionCheckboxes} from "./PortsByRegionCheckboxes";
+import {RccuByRegionAccess} from "./RccuByRegionAccess";
 import _ from "lodash/fp";
 
 
 const Declaration = styled('div')(({theme}) => ({
-  textAlign: "left",
-  padding: theme.spacing(2),
-  width: '100%',
+    textAlign: "left",
+    padding: theme.spacing(2),
+    width: '100%',
 }))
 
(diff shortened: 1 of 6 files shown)


react/src/components/AccessRequestForm.tsx
```diff
@@ -10,151 +10,167 @@ import ListItemText from "@mui/material/ListItemText";
...
+                <ListItem>
+                    <PortsByRegionCheckboxes regions={props.regions} setPorts={setSelectedPorts}
+                                             selectedPorts={selectedPorts}/>
+                </ListItem>
+                <Divider/>
+                <p>Please select the rccu region you require access to</p>
+                <RccuByRegionAccess regions={props.regions} setSelectedRccuRegions={setSelectedRccuRegions}
+                                    selectedRccuRegions={selectedRccuRegions}/>
+                <Divider/>
+                <ListItem
```
richbirch, line 93: I think we need to move the staffing access checkbox up above the RCCU section as it's relevant to the ports rather than RCCU users. I also noticed we could improve the staffing text since it's currently ambiguous as to what it gives people and ends up with users requesting it because they think they need it to view staffing. Can we change it to:
`I require access to enter staffing figures as my role includes planning`

react/src/components/AccessRequestForm.tsx
```diff
@@ -10,151 +10,167 @@ import ListItemText from "@mui/material/ListItemText";
...
+        return <Box sx={{width: '100%'}}>
+            <h1>Welcome to DRT</h1>
+            <p>Please select the ports you require access to</p>
+            <List>
+                <ListItem>
+                    <PortsByRegionCheckboxes regions={props.regions} setPorts={setSelectedPorts}
+                                             selectedPorts={selectedPorts}/>
+                </ListItem>
+                <Divider/>
+                <p>Please select the rccu region you require access to</p>
```
richbirch: Let's upper-case `RCCU`

richbirch (changes requested):
Nice one. A couple of minor changes to make but pretty much there

richbirch (approved):
🚲

PR #72, opened 2022-08-25
Drtii 1038

ngu04:
- Choosing between RCCU and Ports selection and more information collection from user before requesting access for DRT port or regions


react/src/components/AccessRequestForm.tsx
```diff
@@ -41,53 +41,99 @@ interface IProps {
...
     }
 
+    const pageMessage = () => {
+        if (state.rccOption === "rccu")
+            return "Please select the RCCU region you require access to"
+        else
+            return "Please select the ports you require access to"
+    }
+
+    const moreInfoConditionCheck = () => {
```
richbirch: `moreInfoConditionCheck` could be more descriptive. For example `moreInfoRequired`

react/src/components/AccessRequestForm.tsx
```diff
@@ -109,57 +155,66 @@ export default function AccessRequestForm(props: IProps) {
...
+                    <ListItemIcon>
+                        <Checkbox
+                            inputProps={{'aria-labelledby': "agreeDeclaration"}}
+                            name="agreeDeclaration"
+                            checked={state.agreeDeclaration}
+                        />
+                    </ListItemIcon>
+                    <ListItemText id="agreeDeclaration" primary="I understand and agree with the above declarations"/>
+                </ListItem>
+                {moreInfoConditionCheck() && state.lineManager.length > 4 && state.agreeDeclaration ?
```
richbirch: We might want to add an email address validation function to the line manager field, and if it's not valid we should display an appropriate message next to the text input so that the user knows why the submit button isn't enabled

richbirch: We should display a disabled submit button until they've met the form requirements rather than hide it

react/src/components/AccessRequestForm.tsx
```diff
@@ -109,57 +155,66 @@ export default function AccessRequestForm(props: IProps) {
...
+                    /> :
+                    <Button
+                        disabled={!(((selectedPorts.length === 1 && state.rccOption === "port") ||
+                            (selectedRegions.length === 1 && state.rccOption === "rccu")) &&
+                            state.agreeDeclaration && !state.staffing)}
+                        onClick={save}
+                        variant="contained"
+                        color="primary"
+                    >
+                        Request access
```
richbirch: Can we move the submit button out of `AccessRequestFormModal` so that we only have one button? The modal form can be triggered when the button is clicked depended on whether more info it needed or not

react/src/components/AccessRequestFormModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+    ports: string[];
+    manageStaff: boolean;
+    portOrRegionText: string;
+    setPortOrRegionText: ((value: (((prevState: string) => string) | string)) => void)
+    staffText: string;
+    setStaffText: ((value: (((prevState: string) => string) | string)) => void)
+    saveCallback: () => void;
+}
+
+export default function AccessRequestFormModal(props: IProps) {
```
richbirch: `AccessRequestFormModal` might be clearer as something like `AccessRequestAdditionalInformationForm`

react/src/components/PortsByRegionCheckboxes.tsx
```diff
@@ -7,104 +7,119 @@ import ListItemText from "@mui/material/ListItemText";
...
+    const allPorts = props.regions.map(r => r.ports).reduce((a, b) => a.concat(b))
+    const allPortsCount = props.regions.reduce((a, b) => a + b.ports.length, 0)
+    const allPortsSelected = props.selectedPorts.length === allPortsCount
+    const allRegions = props.regions.map(r => r.name)
 
-          function toggleRegionPorts() {
-            if (!regionSelected) props.setPorts(props.selectedPorts.concat(sortedPorts))
-            else props.setPorts(props.selectedPorts.filter(p => !sortedPorts.includes(p)))
-          }
+    const updateByRegionSelection = (region: string) => {
```
richbirch: `updateByRegionSelection` - remove the 'By' -> `updateRegionSelection`

src/main/scala/uk/gov/homeoffice/drt/notifications/EmailNotifications.scala
```diff
@@ -11,32 +11,59 @@ import scala.util.Try
...
       "lineManager" -> manager,
-      "agreeDeclaration" -> agreeDeclaration).asJava
+      "agreeDeclaration" -> agreeDeclaration,
+      "portOrRegionText" -> getTextForField(accessRequest.portOrRegionText),
+      "staffText" -> getTextForField(accessRequest.staffText)).asJava
 
     accessRequestEmails.map { accessRequestEmail =>
-      val maybeResponse = Try(client.sendEmail(
+      val maybeResponse: Try[SendEmailResponse] = Try(client.sendEmail(
         "5f34d7bb-293f-481c-826b-62661ba8a736",
```
richbirch: Would be nice to extract these template ids to vals with a descriptive name now we have more than one, just to make it clear that they're specific templates

richbirch (changes requested):
Nice job! A few bits of tidying but looking good :)

react/src/components/AccessRequestForm.tsx
ngu04: Model can be triggered when user click button but I think once user fill the form in model  they need something to click saying they are done .  Maybe we can rename it 'Request submit' to 'Done' ?

react/src/components/AccessRequestForm.tsx
```diff
@@ -72,22 +98,74 @@ export default function AccessRequestForm(props: IProps) {
...
+
+    const saveOrModal = () => {
+        if (singlePortOrRegion()) {
+            save()
+        } else {
+            setOpenModal(true);
+        }
+    }
+
+    const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
richbirch: `handleChange` is quite vague. How about `handleEmailChange`

richbirch (approved):
:boom:

PR #73, opened 2022-08-30
DRTII-1038 adding condition for linemanager email

ngu04:
- adding condition for line manager email at server side as well .


richbirch (approved):
:kangaroo:

PR #76, opened 2022-10-10
Drtii 1034 1

ngu04:
User request access as part for user management 
- showing user requested for access 
- approval process 
- showing approval list of user
- user access details on row click


build.sbt
```diff
@@ -45,6 +45,11 @@ lazy val root = (project in file(".")).
       "com.github.tototoshi" %% "scala-csv" % scalaCsv,
       "org.scalactic" %% "scalactic" % scalaTestVersion,
 
+      "com.typesafe.slick" %% "slick" % "3.4.1",
+//      "org.slf4j" % "slf4j-nop" % "1.7.26",
```
richbirch: remove commented line

database_setup/aggregated-user_access_requests.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE public.user_access_requests(
+  email text NOT NULL,
+  all_ports boolean,
+  ports text,
+  regions text,
+  staffing boolean NOT NULL,
```
richbirch: can we rename this to `staff_editing` to allow us to introduce a view-only `staffing` permission later?

database_setup/aggregated-user_access_requests.sql
```diff
@@ -0,0 +1,19 @@
...
+  line_manager text,
+  agree_declaration boolean NOT NULL,
+  rcc_option boolean NOT NULL,
```
richbirch: Could we change this to something like `account_type` where it would be either 'port' or 'rccu'? This way we can easily add 'national' later without adding a new field

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -0,0 +1,191 @@
...
+        headerName: 'Line Manager',
+        width: 200,
+    },
+
+];
+
+export default function UserAccess() {
+    const [userRequestList, setUserRequestList] = React.useState([]);
+    const [rowsData, setRowsData] = React.useState([]);
+    const [count, setCount] = React.useState(0);
```
richbirch: Can we give `count` a meaningful name?

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -0,0 +1,191 @@
...
+        });
+        return item;
+    }
+
+    const rowClickOpen = (userData: UserRequestedAccessData) => {
+        setRowDetails(userData)
+        setOpenModal(true)
+    }
+
+    const addUserToGroups = () => {
```
richbirch: I think the action here should be `approveUserAccessRequest`. Adding the user to various groups is an implementation detail

src/main/scala/uk/gov/homeoffice/drt/routes/ApiRoutes.scala
```diff
@@ -164,6 +183,54 @@ object ApiRoutes extends JsonSupport
             }
           }
         },
+        (get & path("userDetails" / Segment)) { userEmail =>
```
richbirch: can we keep to the 'kebab-case' convention for the end point names rather than introducing camelCase?

src/main/scala/uk/gov/homeoffice/drt/routes/ApiRoutes.scala
```diff
@@ -164,6 +183,54 @@ object ApiRoutes extends JsonSupport
...
+                            KeyCloakUser("", "", false, false, "", "", "")
+                        }
+                    }
+                  complete(keyCloakUser)
+                }
+              }
+            }
+          }
+        },
+        (post & path("addUserToGroup" / Segment)) { id =>
```
richbirch: This end point should be `accept-user-request` as this is the desired behaviour - the endpoint doesn't know anything about groups

src/main/scala/uk/gov/homeoffice/drt/routes/ApiRoutes.scala
```diff
@@ -78,6 +86,17 @@ object ApiRoutes extends JsonSupport
             }
           }
         },
+        (get & path("request-access")) {
```
richbirch: I think this route should be `access-request` as that's what it's returning rather than a `request access` is a very, so suggests it's doing something rather than fetching access request details

src/main/scala/uk/gov/homeoffice/drt/notifications/EmailNotifications.scala
```diff
@@ -15,10 +15,22 @@ case class EmailNotifications(apiKey: String, accessRequestEmails: List[String])
 
   val accessRequestLineManagerNotificationEmailTemplateId = "c80595c3-957a-4310-a419-b2f254df3909"
 
+  val accessGrantedTemplateId = "12e36257-c485-4e13-af4f-2293d2dd34a6"
+
   def getFirstName(email: String): String = {
     Try(email.split("\\.").head.toLowerCase.capitalize).getOrElse(email)
   }
 
+  def sendAccessGranted(requester: String) = {
```
richbirch: can we add a bcc to the drt poise email address so that we can see the emails being sent out to users for a while?

src/main/scala/uk/gov/homeoffice/drt/db/UserAccessRequestDao.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  implicit val userAccessRequestFormatParser: RootJsonFormat[UserAccessRequest] = jsonFormat12(UserAccessRequest)
+}
+
+case class UserAccessRequest(
+  email: String,
+  portsRequested: String,
+  allPorts: Boolean,
+  regionsRequested: String,
+  staffing: Boolean,
```
richbirch: let's rename this to `staff_editing`

src/main/scala/uk/gov/homeoffice/drt/db/UserAccessRequestDao.scala
```diff
@@ -0,0 +1,113 @@
...
+  lineManager: String,
+  agreeDeclaration: Boolean,
+  rccOption: Boolean,
```
richbirch: can this be an `accountType` string, ie 'port' or 'region'

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -0,0 +1,191 @@
...
+        setUserDetails([]);
+        let fond = selectedRows.map(s => findEmail(s))
```
richbirch, line 63: shall we remove the `fond` (typo?) variable and just add a call to keyCloakUserDetails to the first map?

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -0,0 +1,191 @@
...
+export default function UserAccess() {
+    const [userRequestList, setUserRequestList] = React.useState([]);
+    const [rowsData, setRowsData] = React.useState([]);
+    const [count, setCount] = React.useState(0);
+    const [openModal, setOpenModal] = React.useState(false)
+    const [rowDetails, setRowDetails] = React.useState({})
+    const [selectedRows, setSelectedRows] = React.useState([]);
+    const [userDetails, setUserDetails] = React.useState([]);
+    const [requestPosted, setRequestPosted] = React.useState(false)
+    const [approvedUserRequest, setApprovedUserRequest] = React.useState(false)
```
richbirch: We can improve the naming here. `showApprovedRequests` would be clearer as to its purpose

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -0,0 +1,191 @@
...
+        setUserRequestList(response.data as UserRequestedAccessData[])
+        setRowsData(response.data as GridRowModel[])
+    }
+
+    const updateUserDetailsState = (response: AxiosResponse) => {
+        setUserDetails([...userDetails, response.data as KeyCloakUser]);
+    }
+
+    const keyCloakUserDetails = (email) => {
+        console.log('keyCloakUserDetails email ' + ApiClient.userDetailsEndpoint + '/' + email)
```
richbirch: remove debug logging?

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -0,0 +1,191 @@
...
+        axios.get(ApiClient.userDetailsEndpoint + '/' + email)
+            .then(response => updateUserDetailsState(response))
+    }
+
+    const userRequested = () => {
+        setCount(1)
+        console.log(ApiClient.requestAccessEndPoint + '?status=Requested')
```
richbirch: remove debug logging?

react/src/components/UserManagement/UserAccessApproved.tsx
```diff
@@ -0,0 +1,46 @@
...
+    transform: 'translate(-50%, -50%)',
+    width: 600,
+    bgcolor: 'background.paper',
+    border: '2px solid #000',
+    boxShadow: 24,
+    p: 4,
+};
+
+interface IProps {
+    userDetails: KeyCloakUser[];
```
richbirch: We want a single type rather than an array, so `userDetails: KeyCloakUser | undefined`

richbirch: `interface KeyCloakUser` in `UserAccess.tsx` needs an `export` keyword so we can access it here, and the import in this file needs to be: `import {KeyCloakUser} from './UserAccess'`

react/src/components/UserManagement/UserAccessApproved.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export default function UserAccessApproved(props: IProps) {
+
+    const resetRequestPosted = () => {
+        console.log('requested userDetails' + props.userDetails.map(ud => ud.email));
+        window.location.reload(true);
+    }
+
+    return (
+        <div class="flex-container">
```
richbirch: `class` should be `className` here - intellij should show you these issues if you have typescript support

react/src/components/UserManagement/UserAccessApproved.tsx
```diff
@@ -0,0 +1,46 @@
...
+            <div>
+                <Box sx={style}>
+                    <Typography align="center" id="modal-modal-title" variant="h6" component="h2">
+                        User Approved
+                    </Typography>
+                    <br/>
+                    <div align="center">Selected User with email {props.userDetails.map(ud => ud.email + ' ')}is granted
```
richbirch: once you've fixed the `userDetails` type, you can access it by checking it's defined first, eg {props.userDetails && ...}

react/src/components/UserManagement/UserRequestDetails.tsx
```diff
@@ -0,0 +1,134 @@
...
+export default function UserRequestDetails(props: IProps) {
+    const [open, setOpen] = React.useState(true);
+//     const handleOpen = () => setOpen(true);
+    const handleClose = () => {
+        props.setOpenModal(false)
+        setOpen(false)
+    }
+
+    return (
+        <div class="flex-container">
```
richbirch: `class` should be `className`

react/src/components/UserManagement/UserRequestDetails.tsx
```diff
@@ -0,0 +1,134 @@
...
+import Modal from '@mui/material/Modal';
+import Table from '@mui/material/Table';
+import TableBody from '@mui/material/TableBody';
+import TableCell from '@mui/material/TableCell';
+import TableContainer from '@mui/material/TableContainer';
+import TableHead from '@mui/material/TableHead';
+import TableRow from '@mui/material/TableRow';
+import Paper from '@mui/material/Paper';
+
+interface UserRequestedAccessData {
```
richbirch: this needs an `export` keyword so the import in `UserAccess.tsx` will work

react/src/components/UserManagement/UserAccessApprovedList.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const [openModal, setOpenModal] = React.useState(false)
+
+    const updateAccessRequestData = (response: AxiosResponse) => {
+        setUserRequestList(response.data as UserRequestedAccessData[])
+        setRowsData(response.data as GridRowModel[])
+    }
+
+    const userRequested = () => {
+        setCount(1)
+        console.log(ApiClient.requestAccessEndPoint + '?status=Approved')
```
richbirch: remove debug log message

react/src/components/UserManagement/UserAccessApprovedList.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+interface IProps {
+    approvedUserRequest: boolean;
+    setApprovedUserRequest: ((value: (((prevState: boolean) => boolean) | boolean)) => void);
+}
+
+export default function UserAccessApprovedList(props: IProps) {
+    const [userRequestList, setUserRequestList] = React.useState([]);
+    const [rowsData, setRowsData] = React.useState([]);
+    const [count, setCount] = React.useState(0);
```
richbirch: let's improve this name, and maybe make it a boolean, eg `userRequested`

react/src/components/UserManagement/UserAccessApprovedList.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const [rowDetails, setRowDetails] = React.useState({})
+    const [openModal, setOpenModal] = React.useState(false)
+
+    const updateAccessRequestData = (response: AxiosResponse) => {
+        setUserRequestList(response.data as UserRequestedAccessData[])
+        setRowsData(response.data as GridRowModel[])
+    }
+
+    const userRequested = () => {
```
richbirch: I think this should be named `requestUser`

src/main/scala/uk/gov/homeoffice/drt/DashboardClient.scala
```diff
@@ -14,6 +14,22 @@ object DashboardClient {
     Http().singleRequest(HttpRequest(HttpMethods.GET, uri))
   }
 
+  def postWithRolesAndKeycloakToken(uri: String, roles: Iterable[Role], keyCloakToken: String)(implicit system: ClassicActorSystemProvider): Future[HttpResponse] = {
```
richbirch: These two functions are identical apart from the action type. Can we remove the duplication?

richbirch (changes requested):
Good stuff! A bit of tidying and naming improvements. And a lot of typescript fixes to make

src/main/scala/uk/gov/homeoffice/drt/notifications/EmailNotifications.scala
ngu04: couldn't fine bcc feature in gov notify

react/src/components/UserManagement/UserRequestDetails.tsx
```diff
@@ -0,0 +1,175 @@
...
+import {KeyCloakUser} from './UserAccessCommon';
+
+export interface UserRequestedAccessData {
+    agreeDeclaration: boolean;
+    allPorts: boolean;
+    email: string;
+    lineManager: string;
+    portOrRegionText: string;
+    portsRequested: string;
+    account_type: string;
```
richbirch: Can we make this field name format consistent with the others?

react/src/components/UserManagement/UserRequestDetails.tsx
```diff
@@ -0,0 +1,175 @@
...
+    regionsRequested: string;
+    requestTime: string;
+    staffText: string;
+    staff_editing: boolean;
```
richbirch: And this one too

src/main/scala/uk/gov/homeoffice/drt/notifications/EmailNotifications.scala
```diff
@@ -15,10 +14,31 @@ case class EmailNotifications(apiKey: String, accessRequestEmails: List[String])
...
+    else
+      s"https://${curad.portsRequested.trim.toLowerCase()}.$domain/"
+  }
+
+  def sendAccessGranted(clientUserRequestedAccessData: ClientUserRequestedAccessData, domain: String) = {
+    val personalisation: util.Map[String, String] =
+      Map("requesterUsername" -> getFirstName(clientUserRequestedAccessData.email),
+        "link" -> getLink(clientUserRequestedAccessData, domain),
+      ).asJava
+    Try(client.sendEmail(
```
richbirch, line 42: We should probably at least log if this fails

richbirch: If there's no bcc/cc option with govnotify, can we send a second copy to our inbox instead. We might not want them in the long run, but initially it'd be good to see that the emails are going out to users and it gives us visibility in case we want to later improve the email wording

ngu04: I was thinking to do that but was not sure . Now you mention as well , I will do it

react/src/components/RegionalPort.tsx
```diff
@@ -11,13 +11,13 @@ interface IProps {
...
 export const RegionalPort = (props: IProps) => {
-    return <div class="flex-container">
+    return <div className="flex-container">
         {props.user.roles.includes("rcc:" + props.region.toLowerCase()) ?
             <div>
                 <h1> {props.region} region dashboard </h1>
                 <p> This is a new page on DRT. You can download an arrivals export covering all port terminals in
                     this region.</p>
-                <div align="left"><ArrivalExport region={props.region}/></div>
+                <div style={{textAlign:"center"}}><ArrivalExport region={props.region}/></div>
```
richbirch: this is changed from left to center. is that intentional?

richbirch (approved):
:trumpet:

react/src/components/RegionalPort.tsx
ngu04: good shout , I put to left

PR #78, opened 2022-10-17
DRTII-1034 adding broder force with port role while Approval

ngu04:


src/main/scala/uk/gov/homeoffice/drt/routes/ApiRoutes.scala
```diff
@@ -219,6 +219,8 @@ object ApiRoutes extends JsonSupport
                         DashboardClient
                           .userDetailDrtApi(s"${Dashboard.drtUriForPortCode("LHR")}/data/addUserToGroup/$id/$port", user.roles, xAuthToken, "POST")
                       })
+                      DashboardClient
+                        .userDetailDrtApi(s"${Dashboard.drtUriForPortCode("LHR")}/data/addUserToGroup/$id/Border%20Force", user.roles, xAuthToken, "POST")
```
richbirch, line 223: I guess we should introduce relevant groups to drt-lib now, as well as the roles. Then we don't have to use a hard-coded string for the 'Border Force' group and others. Let's merge this now and I'll add a ticket to add groups to drt-lib

richbirch (approved):

PR #81, opened 2022-10-28
Drtii 1099

ngu04:
multiple user access request
dismiss user access
multiple dismiss user access 
email links on mail to open email form 
fixing single click approval on modal
fixing staff access


react/src/components/UserManagement/ConfirmUserAccess.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+interface IProps {
+    emails: string[]
+    message: string
+}
+
+export default function ConfirmUserAccess(props: IProps) {
+
+    const resetRequestPosted = () => {
+        window.location.reload();
```
richbirch: Let's not reload the page. We can just send a request for the latest access requests

react/src/components/UserManagement/ConfirmUserAccess.tsx
```diff
@@ -0,0 +1,73 @@
...
+    }
+    const logMessage = () => {
+        console.log('emails' + props.emails + ' ' + props.emails.map(e => e + ' '))
+    }
+
+    const moreThanOneUserDisplay = () => {
+        return <div>
+            Following selected users are {props.message} access requested.
```
richbirch: ```The following users have had their request {props.message}```

react/src/components/UserManagement/ConfirmUserAccess.tsx
```diff
@@ -0,0 +1,73 @@
...
+                        />
+                    </ListItem>,
+                )}
+            </List>
+        </div>
+    }
+
+    const singleUserDisplay = () => {
+        return <div>
+            Selected User with
```
richbirch: ```{email} has had their request {props.message}```

react/src/components/UserManagement/ConfirmUserAccess.tsx
```diff
@@ -0,0 +1,73 @@
...
+    }
+    const logMessage = () => {
+        console.log('emails' + props.emails + ' ' + props.emails.map(e => e + ' '))
+    }
+
+    const moreThanOneUserDisplay = () => {
+        return <div>
+            Following selected users are {props.message} access requested.
+            <List>
+                {props.emails.filter(e => e != undefined).map(e =>
```
richbirch: How about we store the emails once and pass them to these functions instead of using `filter` 3 times. I'm also not sure why you'd need to check for undefined elements of an array

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,18 +13,22 @@ import Tab from '@mui/material/Tab';
 import {columns, KeyCloakUser} from "./UserAccessCommon";
 
 export default function UserAccess() {
+    const [apiRequestListCount, setApiRequestListCount] = React.useState(0);
```
richbirch: This looks like it should be a boolean and be called something like `accessRequestListRequested`

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,18 +13,22 @@ import Tab from '@mui/material/Tab';
...
     const [apiRequestCount, setApiRequestCount] = React.useState(0);
     const [openModal, setOpenModal] = React.useState(false)
     const [rowDetails, setRowDetails] = React.useState({} as UserRequestedAccessData | undefined)
+    const [selectedRowDetails, setSelectedRowDetails] = React.useState([] as UserRequestedAccessData[]);
     const [selectedRows, setSelectedRows] = React.useState([]);
-    const [userDetails, setUserDetails] = React.useState({} as KeyCloakUser);
+    const [userDetails, setUserDetails] = React.useState([] as KeyCloakUser[]);
     const [requestPosted, setRequestPosted] = React.useState(false)
+    const [dismissedPosted, setDismissedPosted] = React.useState(false)
     const [showApprovedRequests, setShowApprovedUserRequest] = React.useState("Requested")
```
richbirch: This sounds like it's a boolean, but it's not. A clearer name might be `statusFilterValue`

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,18 +13,22 @@ import Tab from '@mui/material/Tab';
...
 
     const handleChange = (event: React.SyntheticEvent, newValue: string) => {
         setShowApprovedUserRequest(newValue);
+        setApiRequestListCount(0);
     };
 
     const getAccessRequest = (response: AxiosResponse) => {
```
richbirch: This name is misleading. It's updating the display, not getting anything

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -33,12 +37,14 @@ export default function UserAccess() {
     }
 
     const updateUserDetailsState = (response: AxiosResponse) => {
-        setUserDetails(response.data as KeyCloakUser);
+        console.log('response' + (response.data as KeyCloakUser).email)
```
richbirch: Do we need these logs?

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,18 +13,22 @@ import Tab from '@mui/material/Tab';
...
+    const [apiRequestListCount, setApiRequestListCount] = React.useState(0);
     const [userRequestList, setUserRequestList] = React.useState([] as UserRequestedAccessData[]);
     const [rowsData, setRowsData] = React.useState([] as GridRowModel[]);
     const [apiRequestCount, setApiRequestCount] = React.useState(0);
     const [openModal, setOpenModal] = React.useState(false)
     const [rowDetails, setRowDetails] = React.useState({} as UserRequestedAccessData | undefined)
+    const [selectedRowDetails, setSelectedRowDetails] = React.useState([] as UserRequestedAccessData[]);
     const [selectedRows, setSelectedRows] = React.useState([]);
-    const [userDetails, setUserDetails] = React.useState({} as KeyCloakUser);
+    const [userDetails, setUserDetails] = React.useState([] as KeyCloakUser[]);
```
richbirch: This should not be an array - you're only using a single user's details

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -33,12 +37,14 @@ export default function UserAccess() {
...
 
-    const keyCloakUserDetails = (email: any) => {
-        axios.get(ApiClient.userDetailsEndpoint + '/' + email)
-            .then(response => updateUserDetailsState(response))
+    const keyCloakUserDetails = (emails: string[]) => {
+        emails.map(email => axios.get(ApiClient.userDetailsEndpoint + '/' + email)
+            .then(response => updateUserDetailsState(response)))
     }
 
     const userRequested = () => {
```
richbirch: This name is also misleading. It sounds like a boolean, but it's getting a list of pending requests. Maybe `fetchPendingRequests`

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,18 +13,22 @@ import Tab from '@mui/material/Tab';
...
 
 export default function UserAccess() {
+    const [apiRequestListCount, setApiRequestListCount] = React.useState(0);
     const [userRequestList, setUserRequestList] = React.useState([] as UserRequestedAccessData[]);
     const [rowsData, setRowsData] = React.useState([] as GridRowModel[]);
     const [apiRequestCount, setApiRequestCount] = React.useState(0);
     const [openModal, setOpenModal] = React.useState(false)
     const [rowDetails, setRowDetails] = React.useState({} as UserRequestedAccessData | undefined)
+    const [selectedRowDetails, setSelectedRowDetails] = React.useState([] as UserRequestedAccessData[]);
     const [selectedRows, setSelectedRows] = React.useState([]);
```
richbirch: This contains the IDs of selected rows, so maybe rename `selectedRowIds`

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -58,32 +64,46 @@ export default function UserAccess() {
...
     }
 
-    const approveUserAccessRequest = () => {
-        setUserDetails({} as KeyCloakUser);
-        let fond = selectedRows.map(s => findEmail(s))
-        console.log('fond hello' + fond.map(a => a?.email));
+    const getKeyClockUserDetails = () => {
+        setUserDetails([{} as KeyCloakUser]);
+        console.log('setUserDetails([{} as KeyCloakUser]) ..' + userDetails.length)
+        let fond: any[] = selectedRows.map(s => findEmail(s))
```
richbirch: How about you get the email from the users and rename this to describe what it contains, eg `emailAddresses`, then you can pass it directly to the keycloak function

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -58,32 +64,46 @@ export default function UserAccess() {
...
+        console.log('fond email' + fond.map(a => a?.email));
         keyCloakUserDetails(fond.map(a => a?.email))
     }
 
+    const addSelectedRowDetails = (srd: any) => {
```
richbirch: Should this type should be `string` rather than `any`?

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -58,32 +64,46 @@ export default function UserAccess() {
...
+        setSelectedRowDetails(oldSelectedRowDetails => [...oldSelectedRowDetails, srd])
+    }
+
     const addSelectedRows = (ids: any) => {
```
richbirch: This parameter type should be `string[]`. Then you won't need to check it if's undefined

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -58,32 +64,46 @@ export default function UserAccess() {
...
         console.log(ids);
+        setSelectedRowDetails([])
+        if (ids) {
+            ids.map((id: string) => addSelectedRowDetails(findEmail(id)))
+        }
         setSelectedRows(ids)
     }
 
     const getRow = (email: string) => {
```
richbirch: A clearer name would be `findRequestByEmail`

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -58,32 +64,46 @@ export default function UserAccess() {
...
         return userRequestList.find(sr => sr.email == email)
     }
 
+    const approveUserAccessRequest = (userDetail: any) => {
+        if (getRow((userDetail as KeyCloakUser).email)) {
```
richbirch: Save the result of `getRow` so we don't call it again on the next line

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,18 +13,22 @@ import Tab from '@mui/material/Tab';
...
     const [userRequestList, setUserRequestList] = React.useState([] as UserRequestedAccessData[]);
     const [rowsData, setRowsData] = React.useState([] as GridRowModel[]);
     const [apiRequestCount, setApiRequestCount] = React.useState(0);
     const [openModal, setOpenModal] = React.useState(false)
     const [rowDetails, setRowDetails] = React.useState({} as UserRequestedAccessData | undefined)
+    const [selectedRowDetails, setSelectedRowDetails] = React.useState([] as UserRequestedAccessData[]);
     const [selectedRows, setSelectedRows] = React.useState([]);
-    const [userDetails, setUserDetails] = React.useState({} as KeyCloakUser);
+    const [userDetails, setUserDetails] = React.useState([] as KeyCloakUser[]);
     const [requestPosted, setRequestPosted] = React.useState(false)
```
richbirch, line 26: A clearer name might be `requestApprovalSent`

react/src/components/UserManagement/UserRequestDetails.tsx
```diff
@@ -61,26 +60,33 @@ export default function UserRequestDetails(props: IProps) {
         setOpen(false)
     }
 
+    const updateState = (keyCloakUser: KeyCloakUser) => {
+        setApiRequestCount(1)
```
richbirch: `apiRequestCount` is not being used apart from in some debugging, so can be removed

react/src/components/UserManagement/UserRequestDetails.tsx
```diff
@@ -61,26 +60,33 @@ export default function UserRequestDetails(props: IProps) {
...
                 .then(response => console.log("User addUserToGroupEndpoint" + response.data))
                 .then(() => setRequestPosted(true))
+                .then(() => setApiRequestCount(0))
         }
-
-
     }, [userDetails]);
 
     const showApprovedButton = () => {
         return !props.approvedPage ?
```
richbirch: `approvedPage` would be clearer as `showApproveButton`

react/src/components/UserManagement/UserRequestDetails.tsx
```diff
@@ -61,26 +60,33 @@ export default function UserRequestDetails(props: IProps) {
...
+    }
+
     const keyCloakUserDetails = () => {
+        console.log('keyCloakUserDetails props.rowDetails?.email ' + props.rowDetails?.email)
         axios.get(ApiClient.userDetailsEndpoint + '/' + props.rowDetails?.email)
-            .then(response => setUserDetails(response.data as KeyCloakUser))
-            .then(() => setApiRequestCount(1))
+            .then(response => updateState(response.data as KeyCloakUser))
+        console.log('keyCloakUserDetails setApiRequestCount ' + apiRequestCount)
+        console.log('keyCloakUserDetails setUserDetails ' + (userDetails as KeyCloakUser))
```
richbirch: If you specify the type of `userDetails` where it's defined then we won't need to user `as` where it's being used

react/src/components/UserManagement/UserRequestDetails.tsx
```diff
@@ -61,26 +60,33 @@ export default function UserRequestDetails(props: IProps) {
         setOpen(false)
     }
 
+    const updateState = (keyCloakUser: KeyCloakUser) => {
+        setApiRequestCount(1)
+        setUserDetails(keyCloakUser)
+    }
+
     const keyCloakUserDetails = () => {
+        console.log('keyCloakUserDetails props.rowDetails?.email ' + props.rowDetails?.email)
```
richbirch: do we need these 3 log lines?

react/src/components/UserManagement/UserRequestDetails.tsx
```diff
@@ -101,11 +107,17 @@ export default function UserRequestDetails(props: IProps) {
                                 <TableBody>
                                     <TableRow>
                                         <TableCell>Email</TableCell>
-                                        <TableCell>{props.rowDetails?.email}</TableCell>
+                                        <TableCell>
+                                            <a href={"mailto:" + props.rowDetails?.email + "?Subject=Access%20grant"}
```
richbirch: the subject should probably be `DRT access request` in these links

richbirch (changes requested):
Things to rename, and a few types to specify or tidy

react/src/components/UserManagement/UserRequestDetails.tsx
ngu04: This is being not just used for debugging

richbirch: Oh yes apologies. It should be a boolean

react/src/components/UserManagement/UserAccess.tsx
ngu04: This changes is for multiple select user request to work for approval.

richbirch: Gotcha. Then let's rename it to reflect the fact it's more than one user. Maybe `users` rather than than `userDetails`

react/src/components/UserManagement/ConfirmUserAccess.tsx
ngu04: I can do that but I need pass state from `useraccess` to `confirmUserAccess` and with `back` click to reset the state and make latest request for request  . Also this is generic for dismiss and approval so i need to pass state for both and back .  With reload it automatically goes by to default with latest request .

richbirch: You can do it! We shouldn't ever need to use location.reload in a single page app

react/src/components/UserManagement/UserAccess.tsx
ngu04: onSelectionModelChange is not able take string[]

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,78 +13,105 @@ import Tab from '@mui/material/Tab';
...
+    const [users, setUsers] = React.useState([] as KeyCloakUser[]);
+    const [receivedUsersResponse, setReceivedUsersResponse] = React.useState(false)
     const [requestPosted, setRequestPosted] = React.useState(false)
-    const [showApprovedRequests, setShowApprovedUserRequest] = React.useState("Requested")
+    const [dismissedPosted, setDismissedPosted] = React.useState(false)
+    const [statusFilterValue, setStatusFilterValue] = React.useState("Requested")
 
     const handleChange = (event: React.SyntheticEvent, newValue: string) => {
-        setShowApprovedUserRequest(newValue);
+        console.log("inside handle change....")
```
richbirch: Probably don't need this log line?

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,78 +13,105 @@ import Tab from '@mui/material/Tab';
...
 
-    const keyCloakUserDetails = (email: any) => {
-        axios.get(ApiClient.userDetailsEndpoint + '/' + email)
+    const getKeyCloakUserDetails = (emails: string[]) => {
+        emails.map(email => axios.get(ApiClient.userDetailsEndpoint + '/' + email)
             .then(response => updateUserDetailsState(response))
+            .then(() => setReceivedUsersResponse(true)))
     }
 
     const userRequested = () => {
```
richbirch: This should be called something like `requestAccessRequests` to show what it's doing

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,78 +13,105 @@ import Tab from '@mui/material/Tab';
...
     const handleChange = (event: React.SyntheticEvent, newValue: string) => {
-        setShowApprovedUserRequest(newValue);
+        console.log("inside handle change....")
+        setApiRequested(false)
+        setStatusFilterValue(newValue);
+        setAccessRequestListRequested(false);
     };
 
-    const getAccessRequest = (response: AxiosResponse) => {
+    const setUserAccessData = (response: AxiosResponse) => {
```
richbirch: This is about access requests, so maybe call it `handleAccessRequestsResponse`

react/src/components/UserManagement/ConfirmUserAccess.tsx
```diff
@@ -0,0 +1,95 @@
...
+    const singleUserDisplay = () => {
+        return <div>
+            {email()} has had their request {messageDisplay()}
+        </div>
+    }
+
+    const messageDisplay = () => {
+        switch (props.message.toLowerCase()) {
+            case "granted" :
+                return "Approved"
```
richbirch: These should all be lowercase as that'll fit both places they're being used

react/src/components/UserManagement/ConfirmUserAccess.tsx
```diff
@@ -0,0 +1,95 @@
...
+    setParentRequestPosted: ((value: (((prevState: boolean) => boolean) | boolean)) => void);
+    apiRequested: boolean
+    setApiRequested: ((value: (((prevState: boolean) => boolean) | boolean)) => void);
+    openModel: boolean
+    setOpenModel: ((value: (((prevState: boolean) => boolean) | boolean)) => void);
+}
+
+export default function ConfirmUserAccess(props: IProps) {
+
+    const email = () => {
```
richbirch: We shouldn't need to check for undefined values here

react/src/components/UserManagement/UserAccess.tsx
richbirch: This name should still be changed to clarify what it's value represents

react/src/components/UserManagement/UserAccessApprovedList.tsx
```diff
@@ -25,8 +27,8 @@ export default function UserAccessApprovedList(props: IProps) {
     }
 
     const approvedAccessApi = () => {
```
richbirch: This would be better as `requestAccessRequestsWithStatus`, and pass the status in as a parameter

react/src/components/UserManagement/UserRequestDetails.tsx
```diff
@@ -47,48 +46,74 @@ interface IProps {
...
 }
 
 export default function UserRequestDetails(props: IProps) {
-    const [open, setOpen] = React.useState(true);
     const [requestPosted, setRequestPosted] = React.useState(false)
-    const [userDetails, setUserDetails] = React.useState({});
-    const [apiRequestCount, setApiRequestCount] = React.useState(0);
-
+    const [user, setUser] = React.useState({} as KeyCloakUser);
+    const [apiRequested, setApiRequested] = React.useState(false);
```
richbirch: I'm struggling to figure out what `apiRequested` represents. I think it really means that some user details have been updated and you want to make another request - it's only set to true after you've received a response to a call for user details. Whatever it means, it really needs a new name to show what its intended effect is

react/src/components/UserManagement/UserRequestDetails.tsx
```diff
@@ -47,48 +46,74 @@ interface IProps {
...
                 .then(response => console.log("User addUserToGroupEndpoint" + response.data))
                 .then(() => setRequestPosted(true))
+                .then(() => setApiRequested(false))
         }
+        if (!props.apiParentRequested) {
+            props.setOpenModal(false)
+        }
+    }, [user, apiRequested]);
+
+    const approvedPage = () => {
```
richbirch: Why is this function called `approvedPage`? It returns a button!

react/src/components/UserManagement/UserRequestDetails.tsx
```diff
@@ -47,48 +46,74 @@ interface IProps {
     openModal: boolean;
     setOpenModal: ((value: (((prevState: boolean) => boolean) | boolean)) => void);
     rowDetails: UserRequestedAccessData | undefined
-    approvedPage: boolean;
+    approvedPage: string;
```
richbirch: `approvedPage` is not a good name. It appears to hold the status of access requests that should be shown

richbirch (changes requested):
Still a lot of name changes needed here. There are a lot of variable and functions with names that have nothing to do with what they're being used for which makes it really difficult to see what's going on.
I'll also submit a PR against this branch to show how to get rid of the `any` types

react/src/components/UserManagement/UserRequestDetails.tsx
ngu04: apiRequested (api request made to get user details) is state to so that user details are called and updated through api in user effect . It may get confusing bcos we pass it to other component and the component can reset it so that api is called again after any update is done by the component for user access request . 
how is `userApiRequested` . Let me know if that make sense .  we can have chat if you prefer it that way

richbirch: If its purpose is to signal that the user should be added to groups then it could be called `readyToAssignGroupsToUser`. Hopefully you see what I mean - `userApiRequested` doesn't really say what it signals (what is the user api that has been requested?)

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,78 +13,108 @@ import Tab from '@mui/material/Tab';
...
 
-    const findEmail = (requestTime: any) => {
+    const findAccessRequestByEmail = (requestTime: string | number) => {
         return userRequestList.find(obj => {
             return obj.requestTime.trim() == requestTime
         });
     }
 
     const rowClickOpen = (userData: UserRequestedAccessData | undefined) => {
+        console.log('rowClickOpen' + userData + ' ' + openModal)
```
richbirch: We probably don't need these logs now?

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,78 +13,108 @@ import Tab from '@mui/material/Tab';
...
         setOpenModal(true)
+        console.log('rowClickOpen after' + userData + ' ' + openModal)
     }
 
-    const approveUserAccessRequest = () => {
-        setUserDetails({} as KeyCloakUser);
-        let fond = selectedRows.map(s => findEmail(s))
-        console.log('fond hello' + fond.map(a => a?.email));
-        keyCloakUserDetails(fond.map(a => a?.email))
+    const approveUserDetails = () => {
```
richbirch: This doesn't approve user details. It fetches user details from keycloak. So maybe name it `fetchUsers`?

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,78 +13,108 @@ import Tab from '@mui/material/Tab';
...
 
-    React.useEffect(() => {
-        if (apiRequestCount == 0) {
-            userRequested();
-        }
-
-        if(getRow((userDetails as KeyCloakUser).email)) {
-            axios.post(ApiClient.addUserToGroupEndpoint + '/' + (userDetails as KeyCloakUser).id, getRow((userDetails as KeyCloakUser).email))
+    const approveUserAccessRequest = (user: KeyCloakUser) => {
+        let email = findRequestByEmail((user as KeyCloakUser).email)
```
richbirch: Good to use `const` if we don't want a mutable variable

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,78 +13,108 @@ import Tab from '@mui/material/Tab';
...
+                .then(() => setReceivedUsersResponse(false))
         }
-    }, [userDetails]);
+    }
+
+    React.useEffect(() => {
+        if (!receivedUserDetails) {
+            requestAccessRequests();
+        }
+        console.log('React.useEffect user details ' + users.length + ' ' + users.map(ud => ud.email))
```
richbirch: Do we want this log?

react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -13,78 +13,108 @@ import Tab from '@mui/material/Tab';
...
+        if (receivedUsersResponse) {
+            users.map(approveUserAccessRequest)
+        }
+    }, [users, receivedUserDetails, requestPosted, receivedUsersResponse]);
```
richbirch: It looks like you can remove `requestPosted` from the dependencies

react/src/components/UserManagement/UserRequestDetails.tsx
```diff
@@ -47,48 +46,72 @@ interface IProps {
...
+    }
+
+    const updateState = (keyCloakUser: KeyCloakUser) => {
+        setReceivedUserDetails(true)
+        setUser(keyCloakUser)
     }
 
     const keyCloakUserDetails = () => {
+        setMessage("Granted")
+        console.log('keyCloakUserDetails props.rowDetails?.email ' + props.rowDetails?.email)
```
richbirch: Do you still want this log?

richbirch (changes requested):
Nearly there! Just a couple more improvements and probably removing a bunch of logs

richbirch (approved):
Nice one!

PR #85, opened 2022-12-02
Drtii 1139 region checkbox logic is inverted

richbirch:
Cleaned up logic for selected ports, regions and all ports
Use checkboxes with labels rather than buttons
Clean up access form state

### DIFF
--- a/src/main/scala/uk/gov/homeoffice/drt/authentication/AccessRequest.scala
+++ b/src/main/scala/uk/gov/homeoffice/drt/authentication/AccessRequest.scala
@@ -1,18 +1,18 @@
 package uk.gov.homeoffice.drt.authentication
 
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport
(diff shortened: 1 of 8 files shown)


src/main/scala/uk/gov/homeoffice/drt/authentication/AccessRequest.scala
```diff
@@ -1,18 +1,18 @@
 package uk.gov.homeoffice.drt.authentication
 
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport
-import spray.json.{ DefaultJsonProtocol, RootJsonFormat }
+import spray.json.{DefaultJsonProtocol, RootJsonFormat}
 
-case class AccessRequest(
-  portsRequested: Set[String],
```
ngu04, line 7: I prefer this indentation but default format make is other way .

ngu04 (approved):
only thing I found is All Regions, North, South , Central , Heathrow not appearing in Bold .

Nice one !

PR #87, opened 2022-12-05
Drtii 1035

ngu04:
- Adding user tracking notification 
 - User tracking revoke access 
 - Akka update
 - Scheduler for notification and revoke access
 - Move keycloak  and using API only related to user tracking at moment 

Note: Later in next ticket, need to move all api that DRT use for Dashboard App back here .

### DIFF
--- a/react/src/components/users/UserTracking.tsx
+++ b/react/src/components/users/UserTracking.tsx
@@ -1,0 +1,89 @@
+import Box from "@mui/material/Box";
+import {DataGrid, GridColDef, GridRowModel} from "@mui/x-data-grid";
+import * as React from "react";
+import axios, {AxiosResponse} from "axios";
+import ApiClient from "../../services/ApiClient";
+import {GridValueFormatterParams} from "@mui/x-data-grid/models/params/gridCellParams";
+import moment from "moment-timezone";
+
+const formatDate = (param:GridValueFormatterParams) => {
+    return moment(param?.value).format("YYYY-MM-DD HH:mm")}
+
+export const userColumns: GridColDef[] = [
+    {
+        field: 'id',
+        headerName: 'Id',
+        width: 100,
+        hide: true
+    },
+    {
+        field: 'username',
+        headerName: 'Username',
+        width: 150,
+    },
+    {
+        field: 'email',
+        headerName: 'Email',
+        width: 280
+    },
+    {
+        field: 'latest_login',
+        headerName: 'Latest Login',
+        width: 200,
+        valueFormatter: params =>
+            params.value ? formatDate(params) : '',
+    },
+    {
+        field: 'inactive_email_sent',
+        headerName: 'Inactive Email Sent',
+        description: 'This column has a value getter and is not sortable.',
+        sortable: false,
+        width: 200,
+        valueFormatter: params =>
+            params.value ? formatDate(params) : '',
+    },
+    {
+        field: 'revoked_access',
+        headerName: 'Revoked Access',
+        width: 200,
+        valueFormatter: params =>
+            params.value ? formatDate(params) : '',
+    },
+
+];
+
+export default function UserTracking() {
+    const [rowsData, setRowsData] = React.useState([] as GridRowModel[]);
+    const [userRequested, setUserRequested] = React.useState(false);
+
+    const handleAccessRequestsResponse = (response: AxiosResponse) => {
+        // setUserRequestList(response.data as UserRequestedAccessData[])
+        setRowsData(response.data as GridRowModel[])
+    }
+
+    const requestAccessRequests = () => {
+        // setReceivedUserDetails(true)
+        axios.get(ApiClient.userEndpoint)
+            .then(response => handleAccessRequestsResponse(response))
+            .then(() => setUserRequested(true))
+    }
+
+    React.useEffect(() => {
+        if (!userRequested) {
+            requestAccessRequests();
+        }
+    });
+
+
+    return (
+        <Box sx={{height: 400, width: '100%'}}>
+            <DataGrid
+                rows={rowsData}
+                columns={userColumns}
+                pageSize={5}
+                rowsPerPageOptions={[5]}
+                experimentalFeatures={{newEditingApi: true}}
+            />
+        </Box>
+    )
+}
--- a/react/src/components/users/UserTracking.tsx
+++ b/react/src/components/users/UserTracking.tsx
@@ -61,16 +61,17 @@ export default function UserTracking() {
     }
 
     const requestAccessRequests = () => {
-        axios.get(ApiClient.userEndpoint)
+        setUserRequested(true)
+        axios.get(ApiClient.userListEndpoint)
             .then(response => handleAccessRequestsResponse(response))
-            .then(() => setUserRequested(true))
+
     }
 
     React.useEffect(() => {
         if (!userRequested) {
             requestAccessRequests();
         }
-    });
+    },[userRequested]);
 
 
     return (
--- a/react/src/components/users/UserTracking.tsx
+++ b/react/src/components/users/UserTracking.tsx
@@ -57,12 +57,10 @@ export default function UserTracking() {
     const [userRequested, setUserRequested] = React.useState(false);
 
     const handleAccessRequestsResponse = (response: AxiosResponse) => {
-        // setUserRequestList(response.data as UserRequestedAccessData[])
         setRowsData(response.data as GridRowModel[])
     }
 
     const requestAccessRequests = () => {
-        // setReceivedUserDetails(true)
         axios.get(ApiClient.userEndpoint)
             .then(response => handleAccessRequestsResponse(response))
             .then(() => setUserRequested(true))
--- a/react/src/services/ApiClient.ts
+++ b/react/src/services/ApiClient.ts
@@ -16,7 +16,7 @@ export default class ApiClient implements IApiClient {
   public static userDetailsEndpoint = "/api/user-details";
   public static addUserToGroupEndpoint = "/api/accept-user-request";
   public static updateUserRequestEndpoint = "/api/update-user-request";
-  public static userEndpoint = "/data/users";
+  public static userListEndpoint = "/data/users";
 
 
   public fetchData(userEndPoint: string, handleResponse: (r: AxiosResponse) => void) {
--- a/react/src/services/ApiClient.ts
+++ b/react/src/services/ApiClient.ts
@@ -16,6 +16,7 @@ export default class ApiClient implements IApiClient {
   public static userDetailsEndpoint = "/api/user-details";
   public static addUserToGroupEndpoint = "/api/accept-user-request";
   public static updateUserRequestEndpoint = "/api/update-user-request";
+  public static userEndpoint = "/data/users";
 
 
   public fetchData(userEndPoint: string, handleResponse: (r: AxiosResponse) => void) {
--- a/react/src/services/ApiClient.ts
+++ b/react/src/services/ApiClient.ts
@@ -16,7 +16,7 @@ export default class ApiClient implements IApiClient {
   public static userDetailsEndpoint = "/api/user-details";
   public static addUserToGroupEndpoint = "/api/accept-user-request";
   public static updateUserRequestEndpoint = "/api/update-user-request";
-  public static userListEndpoint = "/data/users";
+  public static userListEndpoint = "/user/all";
 
 
   public fetchData(userEndPoint: string, handleResponse: (r: AxiosResponse) => void) {
--- a/src/main/scala/uk/gov/homeoffice/drt/Server.scala
+++ b/src/main/scala/uk/gov/homeoffice/drt/Server.scala
@@ -77,7 +77,7 @@ object Server {
       DrtRoutes("drt", serverConfig.portIataCodes),
       ApiRoutes("api", serverConfig.clientConfig, notifications, userRequestService, neboRoutes),
       ExportRoutes(new ProdHttpClient),
-      UserRoutes("data", userService))
+      UserRoutes("user", userService))
     val serverBinding: Future[Http.ServerBinding] = Http().newServerAt(serverConfig.host, serverConfig.port).bind(routes)
 
     ctx.pipeToSelf(serverBinding) {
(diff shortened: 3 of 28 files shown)


react/src/components/users/UserTracking.tsx
```diff
@@ -0,0 +1,87 @@
...
+        setRowsData(response.data as GridRowModel[])
+    }
+
+    const requestAccessRequests = () => {
+        axios.get(ApiClient.userEndpoint)
+            .then(response => handleAccessRequestsResponse(response))
+            .then(() => setUserRequested(true))
+    }
+
+    React.useEffect(() => {
```
richbirch, line 70: This should have a dependencies list as the second argument to `useEffect` which in this case would give `userRequested` as a dependency

react/src/components/users/UserTracking.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const [userRequested, setUserRequested] = React.useState(false);
+
+    const handleAccessRequestsResponse = (response: AxiosResponse) => {
+        setRowsData(response.data as GridRowModel[])
+    }
+
+    const requestAccessRequests = () => {
+        axios.get(ApiClient.userEndpoint)
+            .then(response => handleAccessRequestsResponse(response))
+            .then(() => setUserRequested(true))
```
richbirch: You want to call `setUserRequested(true)` before the call to axios so that `useEffect` doesn't try to execute the api call again before the response has come back

react/src/services/ApiClient.ts
```diff
@@ -16,6 +16,7 @@ export default class ApiClient implements IApiClient {
   public static userDetailsEndpoint = "/api/user-details";
   public static addUserToGroupEndpoint = "/api/accept-user-request";
   public static updateUserRequestEndpoint = "/api/update-user-request";
+  public static userEndpoint = "/data/users";
```
richbirch: This needs a better name. There's already `userEndPoint` at the top of this list so the only difference is a lower case 'p'. Is there a reason it's prefixed with `/data` rather than `/api` like all the existing end points?

src/main/scala/uk/gov/homeoffice/drt/Server.scala
```diff
@@ -56,8 +75,9 @@ object Server {
         staticResourceDirectory = getFromResourceDirectory("frontend/static")).route,
       CiriumRoutes("cirium", serverConfig.ciriumDataUri),
       DrtRoutes("drt", serverConfig.portIataCodes),
-      ApiRoutes("api", serverConfig.clientConfig, notifications, neboRoutes),
-      ExportRoutes(new ProdHttpClient))
+      ApiRoutes("api", serverConfig.clientConfig, notifications, userRequestService, neboRoutes),
+      ExportRoutes(new ProdHttpClient),
+      UserRoutes("data", userService))
```
richbirch: I think the user routes belong under the existing `/api` prefix - so pass it to `ApiRoutes`

src/main/scala/uk/gov/homeoffice/drt/db/AppDatabase.scala
```diff
@@ -0,0 +1,15 @@
...
+import slick.lifted.TableQuery
+import slick.jdbc.PostgresProfile.api._
+
+object AppDatabase {
+
+  lazy val db: PostgresProfile.backend.Database = Database.forConfig("postgresDB")
+
+  val userTable = TableQuery[UserTable]
+
+  val userAccessRequests: TableQuery[UserAccessRequestsTable] = TableQuery[UserAccessRequestsTable]
```
richbirch: To match the naming convention of `userTable` this should have `Table` appended

src/main/scala/uk/gov/homeoffice/drt/db/UserDao.scala
```diff
@@ -0,0 +1,96 @@
...
+  def filterInactive(numberOfInactivityDays: Int)(implicit executionContext: ExecutionContext): Future[Seq[User]] = {
+    db.run(userTable.result)
+      .mapTo[Seq[User]]
+      .map(_.filter(u => u.inactive_email_sent.isEmpty && u.latest_login.toLocalDateTime.isBefore(LocalDateTime.now().minusDays(numberOfInactivityDays))))
+  }
+
+  def filterUserToRevoke()(implicit executionContext: ExecutionContext): Future[Seq[User]] = {
+    db.run(userTable.result)
+      .mapTo[Seq[User]]
+      .map(_.filter(u => u.revoked_access.isEmpty && u.inactive_email_sent.nonEmpty && u.inactive_email_sent.exists(_.toLocalDateTime.isBefore(LocalDateTime.now().minusDays(7)))))
```
richbirch: Yo can remove `u.inactive_email_sent.nonEmpty &&` as it's made redundant by the next condition

react/src/services/ApiClient.ts
ngu04: I am thinking to move all endpoint for keyclock to data .  so all the `/api` endpoints which are calling DRT app for keycloak , I am thinking to move the under `/data` endpoint

src/main/scala/uk/gov/homeoffice/drt/keycloak/KeycloakService.scala
```diff
@@ -0,0 +1,72 @@
...
+import uk.gov.homeoffice.drt.authentication.KeyCloakUser
+import uk.gov.homeoffice.drt.http.ProdSendAndReceive
+
+import java.time.Instant
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.{ Await, ExecutionContext, Future }
+
+case class TokenData(username: String, creationTime: Long, keyCloakAuthToken: KeyCloakAuthToken)
+
+object KeycloakService {
```
richbirch: It feels like this should be an actor as it's managing some state asynchronously

src/main/scala/uk/gov/homeoffice/drt/keycloak/KeycloakService.scala
```diff
@@ -0,0 +1,72 @@
...
+  var manageUserToken: Option[TokenData] = None
+
+  def isTokenExpired(tokenData: TokenData): Boolean = {
+    Instant.now().getEpochSecond - (tokenData.creationTime + tokenData.keyCloakAuthToken.expiresIn) > 0
+  }
+
+  def getTokenFromFuture(keyCloakAuthToken: Future[KeyCloakAuthResponse]): KeyCloakAuthToken = {
+    Await.result(keyCloakAuthToken, 2.seconds).asInstanceOf[KeyCloakAuthToken]
```
richbirch: We won't need this Await if this class becomes an actor

src/main/scala/uk/gov/homeoffice/drt/keycloak/KeycloakService.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  def getManageUserToken(keyClockConfig: KeyClockConfig, manageUsername: String, managePassword: String)(implicit actorSystem: ActorSystem): KeyCloakAuthToken = {
+    manageUserToken match {
+      case Some(tokenData) => if (isTokenExpired(tokenData)) {
+        val keyCloakAuthToken = getTokenFromFuture(getUserToken(keyClockConfig, manageUsername, managePassword))
+        manageUserToken = Some(TokenData(manageUsername, Instant.now().getEpochSecond, keyCloakAuthToken))
+        keyCloakAuthToken
+      } else tokenData.keyCloakAuthToken
+      case None =>
+        val keyCloakAuthToken = getTokenFromFuture(getUserToken(keyClockConfig, manageUsername, managePassword))
```
richbirch: These 3 lines are exactly the same as the 3 in the other case above

src/main/scala/uk/gov/homeoffice/drt/notifications/EmailNotifications.scala
```diff
@@ -107,4 +111,23 @@ case class EmailNotifications(apiKey: String, accessRequestEmails: List[String])
 
     }
   }
+
+  def sendUserEmailNotification(email: String, domain: String, teamEmail: String, templateId: String, reference: String): Try[SendEmailResponse] = {
```
richbirch: This should either be named to make it clear that it's for 2 specific templates, or we create a type for the template so that only 2 templates that work with this function can be passed to it. At the moment the name suggests you can send any user email you like

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTrackingBehavior.scala
```diff
@@ -0,0 +1,117 @@
...
+
+class UserTracking(
+  serverConfig: ServerConfig,
+  notifications: EmailNotifications,
+  userService: UserService,
+  timers: TimerScheduler[UserTracking.Command],
+  after: FiniteDuration,
+  frequency: Int,
+  numberOfInactivityDays: Int,
+  userTrackingFeatureFlag: Boolean,
```
richbirch: I'm trying to figure out what this parameter does. It looks to me like it controls whether this actor does anything or not. If that's the case then we should rather decide whether to start the actor or not outside of this class. Is there a case for when we want to start the actor but not the timers?

src/main/scala/uk/gov/homeoffice/drt/services/UserService.scala
```diff
@@ -0,0 +1,24 @@
+package uk.gov.homeoffice.drt.services
+
+import uk.gov.homeoffice.drt.db.{ IUserDao, User, UserDao }
+
+import scala.concurrent.ExecutionContext
+
+class UserService(userDao: IUserDao) {
+  def getUser()(implicit ec: ExecutionContext) = {
```
richbirch: This is returning all users, not a single user

src/main/scala/uk/gov/homeoffice/drt/services/UserService.scala
```diff
@@ -0,0 +1,24 @@
...
+    userDao.selectAll
+  }
+
+  def inactiveUserCheck(numberOfInactivityDays: Int)(implicit ec: ExecutionContext) = {
```
richbirch: What does this function return? Its name doesn't tell me what it does

src/main/scala/uk/gov/homeoffice/drt/services/UserService.scala
```diff
@@ -0,0 +1,24 @@
...
+    userDao.filterInactive(numberOfInactivityDays)
+  }
+
+  def getUserToRevoke()(implicit ec: ExecutionContext) = {
```
richbirch: This should be `getUsersToRevoke` as it returns a collection

src/test/scala/uk/gov/homeoffice/drt/db/MockUserDao.scala
```diff
@@ -0,0 +1,22 @@
...
+
+class MockUserDao() extends IUserDao {
+  var userList = Seq.empty[User]
+
+  override def insertOrUpdate(userData: User): Future[Int] = {
+    userList = userList :+ userData
+    Future.successful(1)
+  }
+
+  override def filterInactive(numberOfInactivityDays: Int)(implicit executionContext: ExecutionContext): Future[Seq[User]] = Future.successful(userList)
```
richbirch: `selectInactiveUsers`

src/test/scala/uk/gov/homeoffice/drt/db/MockUserDao.scala
```diff
@@ -0,0 +1,22 @@
...
+
+  override def filterUserToRevoke()(implicit executionContext: ExecutionContext): Future[Seq[User]] = {
```
richbirch: `selectUsersToRevokeAccess`

src/test/scala/uk/gov/homeoffice/drt/routes/UserRoutesSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+    "test2@test.com",
+    new Timestamp(stringToLocalDateTime("2022-12-05T10:15:30.00Z").toEpochMilli),
+    None,
+    None))
+
+  val routes: Route = UserRoutes(
+    "data",
+    userService)
+
+  "Given the users access the drt" >> {
```
richbirch: Can you improve the description of what we're testing here?

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTrackingBehavior.scala
ngu04: We can control at actor level as well . This flag would off in preprod we don't need this run .

src/main/scala/uk/gov/homeoffice/drt/keycloak/KeyCloakAuthTokenService.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  def isTokenExpired(tokenData: TokenData): Boolean = {
+    Instant.now().getEpochSecond - (tokenData.creationTime + tokenData.keyCloakAuthToken.expiresIn) > 0
+  }
+
+  def getTokenFromFuture(manageUsername: String, keyCloakAuthTokenF: Future[KeyCloakAuthResponse])(implicit actorSystem: ActorSystem): Future[KeyCloakAuthToken] = {
+    implicit val ec = actorSystem.dispatcher
+    keyCloakAuthTokenF.map { keyCloakAuthResponse =>
+      val KeyCloakAuthToken = keyCloakAuthResponse.asInstanceOf[KeyCloakAuthToken]
+      manageUserToken = Some(TokenData(manageUsername, Instant.now().getEpochSecond, KeyCloakAuthToken))
```
richbirch: We setting the state in an actor we want to do it while the actor is processing a message to avoid concurrency issues. This is setting it when the future has resolved, so instead you want to send the actor a message to itself to set the value. Eg `context.self ! SetToken(manageUserToken)`, and then add a case to handle that message and actually set the new value

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTrackingBehavior.scala
```diff
@@ -0,0 +1,125 @@
...
+      maxSize, context).userBehaviour)
+  }
+}
+
+class UserTracking(
+  serverConfig: ServerConfig,
+  notifications: EmailNotifications,
+  userService: UserService,
+  timers: TimerScheduler[UserTracking.Command],
+  after: FiniteDuration,
```
richbirch: maybe this could be renamed something like `timerInitialDelay`

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTrackingBehavior.scala
```diff
@@ -0,0 +1,125 @@
...
+  frequency: Int,
```
richbirch: We can pass a duration here rather than an Int. Then it can be named `timerInterval`

richbirch (approved):
Great improvements here!
A couple of minor tweaks and we're good to go :+1:

PR #91, opened 2022-12-15
Drtii 1151

ngu04:
- moving apis from `api` to  `user` for user tracking and user access 
- Moving all actors to typed actors
- Adding user route unit test
- refactor code for testing

### DIFF
--- a/react/src/services/ApiClient.ts
+++ b/react/src/services/ApiClient.ts
@@ -8,14 +8,14 @@ interface IApiClient {
 export default class ApiClient implements IApiClient {
   public static userEndPoint = "/api/user";
   public static configEndPoint = "/api/config";
-  public static requestAccessEndPoint = "/api/access-request";
   public static logoutEndPoint = "/oauth/logout";
   public static alertsEndPoint = "/api/alerts";
   public static neboFileUpload = "/api/nebo-upload";
   public static redListUpdates = "/api/red-list-updates";
-  public static userDetailsEndpoint = "/api/user-details";
-  public static addUserToGroupEndpoint = "/api/accept-user-request";
-  public static updateUserRequestEndpoint = "/api/update-user-request";
+  public static requestAccessEndPoint = "/user/access-request";
+  public static userDetailsEndpoint = "/user/user-details";
+  public static addUserToGroupEndpoint = "/user/accept-user-request";
+  public static updateUserRequestEndpoint = "/user/update-user-request";
   public static userListEndpoint = "/user/all";
 
 
@@ -28,7 +28,7 @@ export default class ApiClient implements IApiClient {
 
   public sendData(userEndPoint: string, data: any, handleResponse: (r: AxiosResponse) => void) {
     axios
-      .post("/api/access-request", data)
+      .post("/user/access-request", data)
       .then(r => handleResponse(r))
       .catch(t => console.log('caught: ' + t))
   }
--- a/react/src/services/ApiClient.ts
+++ b/react/src/services/ApiClient.ts
@@ -12,10 +12,10 @@ export default class ApiClient implements IApiClient {
   public static alertsEndPoint = "/api/alerts";
   public static neboFileUpload = "/api/nebo-upload";
   public static redListUpdates = "/api/red-list-updates";
-  public static requestAccessEndPoint = "/user/access-request";
   public static userDetailsEndpoint = "/user/user-details";
-  public static addUserToGroupEndpoint = "/user/accept-user-request";
-  public static updateUserRequestEndpoint = "/user/update-user-request";
+  public static requestAccessEndPoint = "/user/access-request";
+  public static addUserToGroupEndpoint = "/user/accept-access-request";
+  public static updateUserRequestEndpoint = "/user/update-access-request";
   public static userListEndpoint = "/user/all";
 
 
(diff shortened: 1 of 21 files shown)


react/src/services/ApiClient.ts
```diff
@@ -8,14 +8,14 @@ interface IApiClient {
...
   public static configEndPoint = "/api/config";
-  public static requestAccessEndPoint = "/api/access-request";
   public static logoutEndPoint = "/oauth/logout";
   public static alertsEndPoint = "/api/alerts";
   public static neboFileUpload = "/api/nebo-upload";
   public static redListUpdates = "/api/red-list-updates";
-  public static userDetailsEndpoint = "/api/user-details";
-  public static addUserToGroupEndpoint = "/api/accept-user-request";
-  public static updateUserRequestEndpoint = "/api/update-user-request";
+  public static requestAccessEndPoint = "/user/access-request";
```
richbirch: Let's keep the naming consistent, so we use `access-request` throughout rather than mixing with `user-request`:
```
access-request
accept-access-request
update-access-request
```

src/main/scala/uk/gov/homeoffice/drt/DrtDashboardApp.scala
```diff
@@ -32,7 +34,7 @@ object DrtDashboardApp extends App {
 
   val system: ActorSystem[Server.Message] = ActorSystem(Server(serverConfig), "DrtDashboard")
   if (serverConfig.userTrackingFeatureFlag) {
-    ActorSystem(UserTracking(serverConfig, 1.minutes, 100), "UserTrackingTimer")
+    ActorSystem(UserTracking(serverConfig, 1.minutes, 100, EmailNotifications(serverConfig.accessRequestEmails, new NotificationClient(serverConfig.notifyServiceApiKey))), "UserTrackingTimer")
```
richbirch: We're instantiating `EmailNotifications` and `NotificationClient` in two places. Let's move the instantiation to a single place that can then be passed to the 2 places it's being used

src/main/scala/uk/gov/homeoffice/drt/db/DateTimeJsonSupport.scala
```diff
@@ -0,0 +1,28 @@
...
+      case JsString(rawDate) => {
+        try {
+          DateTime.parse(rawDate)
+        } catch {
+          case iae: IllegalArgumentException => deserializationError("Invalid date format")
+          case _: Exception => None
+        }
+      } match {
+        case dateTime: Timestamp => dateTime
+        case None => deserializationError(s"Couldn't parse date time, got $rawDate")
```
richbirch: I don't think we need this `match` - all you're doing is changing the `None` to a `deserialisationError`, which you could have done in the `catch`

src/main/scala/uk/gov/homeoffice/drt/keycloak/KeycloakService.scala
```diff
@@ -1,20 +1,51 @@
...
+import org.slf4j.{ Logger, LoggerFactory }
 import uk.gov.homeoffice.drt.authentication.KeyCloakUser
 
 import scala.concurrent.{ ExecutionContext, Future }
 
-class KeycloakService(keycloakClient: KeycloakClient) {
+trait IKeycloakService {
+  def getUsersForEmail(email: String): Future[Option[KeyCloakUser]]
+  def removeUser(userId: String): Future[HttpResponse]
+  def addUserToGroup(userId: String, group: String)(implicit ec: ExecutionContext): Future[HttpResponse]
```
richbirch: If you move the `implicit  ec:...` to be a property of the trait it makes the function definition cleaner and removes the need for an execution context to be available every time the function is called

src/main/scala/uk/gov/homeoffice/drt/keycloak/KeycloakService.scala
```diff
@@ -1,20 +1,51 @@
...
+  def logout(username: String)(implicit ec: ExecutionContext): Future[Option[Future[HttpResponse]]]
```
richbirch: Same here

src/main/scala/uk/gov/homeoffice/drt/keycloak/KeycloakService.scala
```diff
@@ -1,20 +1,51 @@
...
+}
+
+class KeycloakService(keycloakClient: KeycloakClient) extends IKeycloakService {
```
richbirch: Would be nice to make this a `case class` - add the implicit execution context as a curried argument to it

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTrackingBehavior.scala
```diff
@@ -1,28 +1,24 @@
...
-import scala.concurrent.duration.{ DurationInt, FiniteDuration }
+import scala.concurrent.ExecutionContext
 
-object UserTracking {
+sealed trait Command
 
-  sealed trait Command
+import scala.concurrent.duration.{ DurationInt, FiniteDuration }
 
+object UserTracking {
```
richbirch, line 21: Do you want to rename the file to match its class name?

richbirch (changes requested):
Nice work. Just a bit of tidying to do

richbirch (approved):
Excellent!

PR #93, opened 2023-01-16
Drtii 1158

ngu04:
- inactivity user access actions for users with no email address 
- adding test using in memory H2 database


src/main/scala/uk/gov/homeoffice/drt/db/UserDao.scala
```diff
@@ -20,19 +20,19 @@ case class User(
   inactive_email_sent: Option[java.sql.Timestamp],
   revoked_access: Option[java.sql.Timestamp])
 
-class UserTable(tag: Tag) extends Table[User](tag, "user") {
+class UserTable(tag: Tag, tableName: String = "user") extends Table[User](tag, tableName) {
 
-  def id = column[String]("id", O.PrimaryKey)
+  def id = column[String]("ID", O.PrimaryKey)
```
richbirch: Is this going to be compatible with the existing lowercase column names? Postgres is case-sensitive, so I'd assume this won't work with the exiting tables in preprod and prod

src/main/scala/uk/gov/homeoffice/drt/db/UserDao.scala
```diff
@@ -46,6 +46,7 @@ trait IUserDao {
 
   def selectAll()(implicit executionContext: ExecutionContext): Future[Seq[User]]
 
+  def deleteAll()(implicit executionContext: ExecutionContext): Future[Int]
```
richbirch: Should we maybe leave these `deleteAll` functions in mocked versions of these classes so they're not accessible to prod code?

src/main/scala/uk/gov/homeoffice/drt/keycloak/KeycloakService.scala
```diff
@@ -25,6 +25,10 @@ case class KeycloakService(keycloakClient: KeycloakClient)(implicit val ec: Exec
     keycloakClient.getUsersForEmail(email)
   }
 
+  def getUsersForUsername(username: String): Future[Option[KeyCloakUser]] = {
```
richbirch: This function is returning a single user, not multiple users

src/main/scala/uk/gov/homeoffice/drt/keycloak/KeycloakService.scala
```diff
@@ -25,6 +25,10 @@ case class KeycloakService(keycloakClient: KeycloakClient)(implicit val ec: Exec
...
+    keycloakClient.getUsersForUsername(username)
```
richbirch: Same here - should be renamed to remove the `s`

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTracking.scala
```diff
@@ -86,27 +90,41 @@ class UserTracking(serverConfig: ServerConfig,
         Behaviors.same
 
       case KeyCloakToken(token: KeyCloakAuthToken) =>
+        context.log.info("KeyCloakToken-RevokeAccess")
```
richbirch, line 99: I guess we don't need this log line

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTracking.scala
```diff
@@ -86,27 +90,41 @@ class UserTracking(serverConfig: ServerConfig,
         Behaviors.same
 
       case KeyCloakToken(token: KeyCloakAuthToken) =>
```
richbirch: If this actor's primary reason for existence is to revoke access to inactive users then it should probably be renamed to reflect that. At the moment sending a `KeyCloakToken` to this actor results in it sending emails and disabling user accounts, which isn't very obvious given its name.

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTracking.scala
```diff
@@ -86,27 +90,41 @@ class UserTracking(serverConfig: ServerConfig,
         Behaviors.same
```
richbirch, line 96: Rather than sending the actor a `KeyCloakToken` to trigger emails and revoke accounts, it would be nice to use `.ask()` in the `RevokeUserCheck` case and map over the future, sending itself an explicit command like `PerformAccountRevocations(keycloakToken)'. That way the actor behaves in a predictable way

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTracking.scala
```diff
@@ -86,27 +90,41 @@ class UserTracking(serverConfig: ServerConfig,
...
+                      serverConfig.teamEmail,
+                      notifications.revokeAccessTemplateId,
+                      "revoked DRT Access")
+                    userService.upsertUser(utr.copy(revoked_access = Some(new Timestamp(new Date().getTime))))
+                    logger.info(s"User with email ${utr.email} access revoked due to inactivity")
+                  }
+                }
+              }
+            } else {
+              keycloakService.getUsersForUsername(utr.username).map { ud =>
```
richbirch: this map is almost identical to the one above, so you can extract the common lines to a function

src/test/scala/uk/gov/homeoffice/drt/db/MockUserDaoSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  val user5 = User(
+    id = "user5@test.com",
+    username = "user5",
+    email = "",
+    latest_login = new Timestamp(Instant.now().toEpochMilli),
+    inactive_email_sent = None,
+    revoked_access = None)
+
+  "select all" should "give all users" in {
```
richbirch: These test are testing the mock rather than the proper class. You need to move the code in the dao class into a common parent so both the tests and the prod code behave the same

src/test/scala/uk/gov/homeoffice/drt/db/UserDaoSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+import org.specs2.mutable.Specification
+import org.specs2.specification.{AfterEach, BeforeEach}
+
+import java.sql.Timestamp
+import java.time.Instant
+import scala.concurrent.Await
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration.DurationInt
+
+class UserDaoSpec extends Specification with AfterEach with BeforeEach {
```
richbirch, line 13: Are you planning to remove this set of tests as you have a set testing the same behaviour using mock?

src/test/scala/uk/gov/homeoffice/drt/db/MockUserDaoSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  val user2 = User(
+    id = "user2@test.com",
+    username = "user2",
+    email = "user2@test.com",
+    latest_login = new Timestamp(Instant.now().toEpochMilli),
+    inactive_email_sent = None,
+    revoked_access = None)
+
+  val user3 = User(
```
richbirch: Would be nice to name these user variables descriptively so the tests are easier to read, eg `userLastLogin61DaysAgo`

richbirch (changes requested):
Nice - a few changes to make to tidy up and correct what you're testing

src/test/scala/uk/gov/homeoffice/drt/db/UserDaoSpec.scala
ngu04: I am thinking to keep this and remove mock .

src/main/scala/uk/gov/homeoffice/drt/db/UserDao.scala
ngu04: Good point,did testing after H2 test .   I will amend change accordingly  Thanks !

src/main/scala/uk/gov/homeoffice/drt/Server.scala
```diff
@@ -1,47 +1,46 @@
...
-import scala.util.{Failure, Success}
+import scala.concurrent.{ ExecutionContextExecutor, Future }
+import scala.util.{ Failure, Success }
 
 case class KeyClockConfig(
-                           url: String,
-                           tokenUrl: String,
-                           clientId: String,
-                           clientSecret: String)
+  url: String,
```
richbirch: Is there a reason you want to change the indentation style in this file?

src/main/scala/uk/gov/homeoffice/drt/routes/UserRoutes.scala
```diff
@@ -30,12 +30,12 @@ object UserRoutes extends JsonSupport
   val log: Logger = LoggerFactory.getLogger(getClass)
 
   def apply(
-             prefix: String,
-             clientConfig: ClientConfig,
-             userService: UserService,
-             userRequestService: UserRequestService,
-             notifications: EmailNotifications,
-             keyClockUrl: String)(implicit ec: ExecutionContextExecutor, system: ActorSystem[Nothing]): Route = {
+    prefix: String,
```
richbirch: Indentation style change here too. Have you changed something in intellij?

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTracking.scala
```diff
@@ -44,15 +47,16 @@ object UserTracking {
...
-class UserTracking(serverConfig: ServerConfig,
-                   notifications: EmailNotifications,
-                   userService: UserService,
-                   timers: TimerScheduler[Command],
-                   timerInitialDelay: FiniteDuration,
-                   timerInterval: FiniteDuration,
-                   numberOfInactivityDays: Int,
-                   maxSize: Int,
-                   context: ActorContext[Command]) {
+class UserTracking(
```
richbirch, line 50: Indentation again

src/main/scala/uk/gov/homeoffice/drt/keycloak/KeyCloakAuthTokenService.scala
```diff
@@ -32,7 +32,7 @@ object KeyCloakAuthTokenService {
           context.log.info("GetToken by {}", replyTo)
           implicit val ec: ExecutionContextExecutor = context.system.executionContext
           implicit val system: ActorSystem[Nothing] = context.system
-          getManageUserToken(keyClockConfig, manageUsername, managePassword, context).map(KeyCloakToken).map(replyTo ! _)
+          getManageUserToken(keyClockConfig, manageUsername, managePassword, context).map(PerformAccountRevocations).map(replyTo ! _)
```
richbirch: `PerformAccountRevocations` doesn't belong here. The `KeyCloakAuthTokenService` shouldn't know about revoking accounts. It should just return a token in response to the `GetToken` command

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTracking.scala
richbirch: Can we either remove this log line or change it to mention the command being processed?

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTracking.scala
```diff
@@ -70,43 +74,61 @@ class UserTracking(serverConfig: ServerConfig,
...
             userService.upsertUser(user.copy(inactive_email_sent = Some(new Timestamp(new Date().getTime))))
-            logger.info(s"User with email ${user.email} notified due to inactivity")
           })
         Behaviors.same
 
       case RevokeUserCheck =>
-        keycloakServiceBehavior ! GetToken(context.self)
+        implicit val timeout = new Timeout(30 seconds)
+        implicit val scheduler = context.system.scheduler
+        keycloakServiceBehavior.ask(ref => GetToken(ref)).map(token => context.self ! token)
```
richbirch: `GetToken` should reply with a `KeyCloakAuthToken`, which we can then send to our self with the `PerformAccountRevocations` command

src/main/scala/uk/gov/homeoffice/drt/Server.scala
ngu04: this is happened by default on my intellij

src/test/scala/uk/gov/homeoffice/drt/db/MockUserDao.scala
```diff
@@ -10,13 +11,24 @@ class MockUserDao() extends IUserDao {
     Future.successful(1)
   }
 
-  override def selectInactiveUsers(numberOfInactivityDays: Int)(implicit executionContext: ExecutionContext): Future[Seq[User]] = Future.successful(userList)
+  override def selectInactiveUsers(numberOfInactivityDays: Int)(implicit executionContext: ExecutionContext): Future[Seq[User]] =
+    Future.successful(userList)
+      .mapTo[Seq[User]]
+      .map(_.filter(u => u.inactive_email_sent.isEmpty && u.latest_login.toLocalDateTime.isBefore(LocalDateTime.now().minusDays(numberOfInactivityDays))))
```
richbirch: If you're keeping this mock then you need to move the logic that's common to the prod version into the `IUserDao` so that they share it rather than duplicating. If we changed the prod logic at some point then at the moment you'd need to remember to also change it here

richbirch (changes requested):
Getting there mate! Just a few more things to fix

src/main/scala/uk/gov/homeoffice/drt/Server.scala
richbirch: Ok. So we've never used this style of indentation. Can you update your intellij settings to match what we usually do?

src/test/scala/uk/gov/homeoffice/drt/db/MockUserDao.scala
ngu04: I will delete MockUserDao as now we got test covered for it

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTracking.scala
```diff
@@ -70,43 +74,61 @@ class UserTracking(serverConfig: ServerConfig,
...
-                    serverConfig.teamEmail,
-                    notifications.revokeAccessTemplateId,
-                    "revoked DRT Access")
-                  userService.upsertUser(utr.copy(revoked_access = Some(new Timestamp(new Date().getTime))))
-                  logger.info(s"User with email ${utr.email} access revoked due to inactivity")
+          utrOption.map { utr: User =>
+            if (utr.email.nonEmpty) {
+              keycloakService.getUsersForEmail(utr.email).map { ud =>
+                ud.map { uId =>
+                  if (utr.email.toLowerCase.trim == uId.email.toLowerCase.trim) {
```
richbirch: This `if` will always be true

richbirch: oh sorry I misread it

richbirch: would be good to rename these to make it easier to read:
`utr` -> `userToRevoke`
`uId` -> `userFromKeycloak`

src/main/scala/uk/gov/homeoffice/drt/schedule/UserTracking.scala
```diff
@@ -70,43 +74,61 @@ class UserTracking(serverConfig: ServerConfig,
...
-                    uId.email,
-                    serverConfig.rootDomain,
-                    serverConfig.teamEmail,
-                    notifications.revokeAccessTemplateId,
-                    "revoked DRT Access")
-                  userService.upsertUser(utr.copy(revoked_access = Some(new Timestamp(new Date().getTime))))
-                  logger.info(s"User with email ${utr.email} access revoked due to inactivity")
+          utrOption.map { utr: User =>
+            if (utr.email.nonEmpty) {
+              keycloakService.getUsersForEmail(utr.email).map { ud =>
```
richbirch: This is another plural that should be singular `getUserForEmail`

richbirch (approved):
Nice!

PR #101, opened 2023-03-08
Drtii 1155 checks

ngu04:
- adding scalastyle and scapegoat
- Steward config for PR frequency

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,7 +1,6 @@
 ThisBuild / libraryDependencySchemes ++= Seq(
   "org.scala-lang.modules" %% "scala-xml" % VersionScheme.Always
 )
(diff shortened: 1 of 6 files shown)


project/plugins.sbt
```diff
@@ -1 +1,9 @@
-addSbtPlugin("com.github.sbt" % "sbt-native-packager" % "1.9.13")
+ThisBuild / libraryDependencySchemes ++= Seq(
+  "org.scala-lang.modules" %% "scala-xml" % VersionScheme.Always
+)
+addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "1.2.0")
```
richbirch: We don't need the scalajs plugin right?

richbirch (approved):
Nice one - just to remove the scalajs sbt plugin before merging

PR #104, opened 2023-03-15
Drtii 1184

ngu04:
- moving filtering logic to slick query

### DIFF
--- a/src/main/scala/uk/gov/homeoffice/drt/db/UserDao.scala
+++ b/src/main/scala/uk/gov/homeoffice/drt/db/UserDao.scala
@@ -45,7 +45,7 @@ trait IUserDao {
 
   def selectInactiveUsers(numberOfInactivityDays: Int)(implicit executionContext: ExecutionContext): Future[Seq[User]]
 
-  def selectUsersToRevokeAccess(numberOfInactivityDays: Int)(implicit executionContext: ExecutionContext): Future[Seq[User]]
+  def selectUsersToRevokeAccess(numberOfInactivityDays: Int, deactivateAfterWarningDays: Int)(implicit executionContext: ExecutionContext): Future[Seq[User]]
 
   def selectAll()(implicit executionContext: ExecutionContext): Future[Seq[User]]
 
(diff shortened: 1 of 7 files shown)


src/main/scala/uk/gov/homeoffice/drt/db/UserDao.scala
```diff
@@ -51,20 +54,29 @@ trait IUserDao {
 class UserDao(db: Database, userTable: TableQuery[UserTable]) extends IUserDao {
   val log: Logger = LoggerFactory.getLogger(getClass)
 
+  def inactiveUserIndex(numberOfInactivityDays: Int): UserTable => Rep[Boolean] = (user: UserTable) =>
```
richbirch: a clearer name could be `noActivitySinceDays`

src/main/scala/uk/gov/homeoffice/drt/db/UserDao.scala
```diff
@@ -51,20 +54,29 @@ trait IUserDao {
...
+    user.inactive_email_sent.isEmpty &&
+      user.latest_login < new Timestamp(Instant.now().minusSeconds(numberOfInactivityDays * 60 * 60 * 24).toEpochMilli)
+
+  def revokeUserIndex: UserTable => Rep[Boolean] = (user: UserTable) =>
```
richbirch: I think this should also check there's been no activity since X days, where X is [number of days considered inactive] + [number of days after inactive email send we deactivate account], eg 90 + 7

richbirch: A clearer name could be `accessShouldBeRevoked`

richbirch (changes requested):
- We need tests around these queries
- Function name clarity

src/main/scala/uk/gov/homeoffice/drt/db/UserDao.scala
ngu04: There are test around queries if you look at UserDaoSpec .  
Also there is update in table which clear out(make column empty) if there is activity by the user .

src/main/scala/uk/gov/homeoffice/drt/db/UserDao.scala
```diff
@@ -34,37 +36,48 @@ class UserTable(tag: Tag, tableName: String = "user") extends Table[User](tag, t
...
   val log: Logger = LoggerFactory.getLogger(getClass)
+  val secondsInADay = 60 * 60 * 24
+
+  def noActivitySinceDays(numberOfInactivityDays: Int): UserTable => Rep[Boolean] = (user: UserTable) =>
+    user.inactive_email_sent.isEmpty &&
+      user.latest_login < new Timestamp(Instant.now().minusSeconds(numberOfInactivityDays * secondsInADay).toEpochMilli)
+
+  def accessShouldBeRevoked(numberOfInactivityDays: Int): UserTable => Rep[Boolean] = (user: UserTable) =>
+    user.revoked_access.isEmpty &&
+      user.latest_login < new Timestamp(Instant.now().minusSeconds((numberOfInactivityDays + 7) * secondsInADay).toEpochMilli) &&
```
richbirch: Shall we pass the `7` in as a parameter named `deativateAfterWarningDays`?

richbirch (approved):
Good stuff - might just be nice to pass the other currently hard coded 7 days in as a parameter

PR #108, opened 2023-04-06
Drtii 1196

ngu04:
- Adding user tracking for user accessing dashboard

### DIFF
--- a/react/src/App.tsx
+++ b/react/src/App.tsx
@@ -52,7 +52,7 @@ const App = (props: PropsFromReact) => {
 
     const [userTracked, setUserTracked] = React.useState(false);
 
-    const userTracking = () => {
+    const trackUser = () => {
         axios
             .get(ApiClient.userTrackingEndPoint)
             .then(() => {
@@ -63,7 +63,7 @@ const App = (props: PropsFromReact) => {
     }
 
     useEffect(() => {
-        userTracking();
+        trackUser();
     }, [userTracked]);
 
     return (props.user.kind === "SignedInUser" && props.config.kind === "LoadedConfig") ?
--- a/react/src/App.tsx
+++ b/react/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useEffect} from 'react';
 import './App.css';
 import {Home} from './components/Home';
 import Alerts from './components/Alerts/Alerts';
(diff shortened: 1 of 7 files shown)


react/src/App.tsx
```diff
@@ -48,6 +50,22 @@ const App = (props: PropsFromReact) => {
     const currentLocation = window.document.location;
     const logoutLink = "/oauth/logout?redirect=" + currentLocation.toString()
 
+    const [userTracked, setUserTracked] = React.useState(false);
+
+    const userTracking = () => {
```
richbirch: Let's use a verb to name the function more clearly, eg `trackUser`

react/src/App.tsx
```diff
@@ -48,6 +50,22 @@ const App = (props: PropsFromReact) => {
...
+            .then(() => {
+                setUserTracked(true)
+            }).catch(reason => {
+            console.log('Unable to user tracking' + reason);
+        })
+    }
+
+    useEffect(() => {
+        userTracking();
+    }, [userTracked]);
```
richbirch, line 67: I don't believe you need to add `userTracked` as a dependency here since the function doesn't read its value

react/src/services/ApiClient.ts
```diff
@@ -7,6 +7,7 @@ interface IApiClient {
 
 export default class ApiClient implements IApiClient {
   public static userEndPoint = "/api/user";
+  public static userTrackingEndPoint = "/api/userTracking";
```
richbirch: I think the end point, like function names, should be named with a verb so that it's clear what it does. So in this case it would be `/api/track-user`

src/main/scala/uk/gov/homeoffice/drt/routes/ApiRoutes.scala
```diff
@@ -53,6 +57,22 @@ object ApiRoutes extends JsonSupport
             }
           }
         },
+        (get & path("userTracking")) {
+          headerValueByName("X-Auth-Roles") { rolesStr =>
+            headerValueByName("X-Auth-Email") { email =>
+              optionalHeaderValueByName("X-Auth-Username") { usernameOption =>
+                userService.upsertUser(
+                  uk.gov.homeoffice.drt.db.User(usernameOption.getOrElse(email),
```
richbirch: Could be nice to name the parameters in this `User` instantiation so we can see what the values are being used for

react/src/App.tsx
ngu04: Use Effect keeps call user tracking we don't add dependency

richbirch (approved):
Looking good. I would just improve the function and end point names to include a verb for clarity

react/src/App.tsx
richbirch: Yes, my bad

PR #110, opened 2023-04-12
DRTII-1208 fixing promise for all user request

ngu04:
- Fixing promise all for while multiple users are selected
- Also RCC region space need blank instead of `%20`

### DIFF
--- a/react/src/components/UserManagement/UserAccess.tsx
+++ b/react/src/components/UserManagement/UserAccess.tsx
@@ -22,7 +22,6 @@ export default function UserAccess() {
     const [selectedRowDetails, setSelectedRowDetails] = React.useState([] as UserRequestedAccessData[]);
     const [selectedRowIds, setSelectedRowIds] = React.useState<GridRowId[]>([]);
     const [users, setUsers] = React.useState([] as KeyCloakUser[]);
-    const [receivedUsersResponse, setReceivedUsersResponse] = React.useState(false)
     const [requestPosted, setRequestPosted] = React.useState(false)
     const [dismissedPosted, setDismissedPosted] = React.useState(false)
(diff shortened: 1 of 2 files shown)


react/src/components/UserManagement/UserAccess.tsx
```diff
@@ -43,9 +43,9 @@ export default function UserAccess() {
     }
 
     const getKeyCloakUserDetails = (emails: string[]) => {
-        emails.map(email => axios.get(ApiClient.userDetailsEndpoint + '/' + email)
-            .then(response => updateUserDetailsState(response))
-            .then(() => setReceivedUsersResponse(true)))
+        Promise.all(emails.map(email => axios.get(ApiClient.userDetailsEndpoint + '/' + email)
```
richbirch: I think we need to refactor this code to be a little easier to follow.
- `fetchUsers` is actually responsible for approving user requests. You wouldn't expect that function to do this with its current name.
- `getKeyCloakUserDetails` triggers an approval, not explicitly but via the `useEffect`. You would not expect this behaviour both because of the name of the function but also because there's no indication in the function that it will lead to anything being approved
- `dismissUserRequest` dismisses one or more requests, so it should be plural `dismissAccessRequests`

I would suggest renaming `fetchUsers` to something that describes what it does, ie `approveSelectedRequests`, have it iterate over the selected users one by one and approve them one by one in the same function. That way we can:
- remove `useEffect`
- remove `receivedUsersResponse` & `setReceivedUsersResponse`
- make it very clear what's going on
- fix the current bug

richbirch (changes requested):
Having looked at the code here it would really benefit from some refactoring. This would both fix the bug and make the code much easier to follow and understand

richbirch (approved):
Nice work!

PR #135, opened 2023-07-13
Drtii 1246 regional export missing arrivals

richbirch:
Change export download to a request for a download
Exports are created in the background and uploaded to s3
When ready a download link is visible in the front end
When failed the download request shows 'failed'


react/src/components/ArrivalExport.tsx
```diff
@@ -1,56 +1,74 @@
...
+  },
+}));
+
+export interface DialogTitleProps {
+  id: string;
+  children?: React.ReactNode;
+  onClose: () => void;
+}
+
+function BootstrapDialogTitle(props: DialogTitleProps) {
```
ngu04: Bootstrap name sound like you are using Bootstrap js here :-)

src/test/scala/uk/gov/homeoffice/drt/routes/ExportRoutesSpec.scala
```diff
@@ -1,64 +1,75 @@
...
+import scala.concurrent.{ExecutionContextExecutor, Future}
 
 
 class ExportRoutesSpec extends AnyWordSpec with Matchers with ScalatestRouteTest {
+  implicit val typedSystem: ActorSystem[Nothing] = ActorSystem.wrap(system)
   implicit val mat: Materializer = Materializer(system)
   implicit val ec: ExecutionContextExecutor = mat.executionContext
 
-  implicit val streamingSupport: CsvEntityStreamingSupport = EntityStreamingSupport.csv().withContentType(ContentTypes.`text/csv(UTF-8)`)
+  implicit val csvStreaming: CsvEntityStreamingSupport = EntityStreamingSupport.csv()//.withFramingRenderer(Flow[ByteString])
```
ngu04: remove the comments

src/main/scala/uk/gov/homeoffice/drt/json/RegionExportJsonFormats.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  implicit object SDateLikeJsonFormat extends RootJsonFormat[SDateLike] {
+    override def read(json: JsValue): SDateLike = json match {
+      case JsNumber(ts) => SDate(ts.toLong)
+    }
+
+    override def write(obj: SDateLike): JsValue = JsNumber(obj.millisSinceEpoch)
+  }
+
+  implicit object RegionExportJsonFormat extends RootJsonFormat[RegionExport] {
```
ngu04: I think instead of all explicit conversion of fields object.toJson should work in with jsonFormat as mentioned below .  just need {region Export Object}.toJson whenever you need json
implicit val RegionExportFormattter: RootJsonFormat[RegionExport] = jsonFormat6(RegionExport)

ngu04 (approved):
Nice one @richbirch !

react/src/components/ArrivalExport.tsx
richbirch: Yeah good point! I copied it from the MUI site, but I've renamed it now

src/main/scala/uk/gov/homeoffice/drt/json/RegionExportJsonFormats.scala
richbirch: Very good point! I've made the changes, thanks

PR #140, opened 2023-07-21
Drtii 1257 regional export no new line at end of each terminal flights

richbirch:
Fix missing new line between end of one terminal and start of another
Add test case
Unify mock http client


src/main/scala/uk/gov/homeoffice/drt/Server.scala
```diff
@@ -88,7 +88,7 @@ object Server {
         CiriumRoutes("cirium", serverConfig.ciriumDataUri),
         DrtRoutes("drt", serverConfig.portIataCodes),
         ApiRoutes("api", serverConfig.clientConfig, neboRoutes, userService),
-        ExportRoutes(ProdHttpClient, exportUploader.upload, exportDownloader.download, () => SDate.now()),
+        ExportRoutes(ProdHttpClient, exportUploader.upload, exportDownloader.download, () => SDate.now(), ProdDatabase),
```
ngu04: We can make prod/test Database as implicit so it is available where ever it is needed  ?

ngu04 (approved):
Nice one !

src/main/scala/uk/gov/homeoffice/drt/Server.scala
richbirch: Yeah I like that. Done!

PR #148, opened 2023-08-14
DRTII-1283 Update drt-lib & add recovery to failed fetch attempts

richbirch:
Update drt-lib to remove DSA role
Add recovery for failed connections


ngu04 (approved):
👍

PR #154, opened 2023-09-12
Drtii 1254 download manager endpoints

richbirch:
- Move regional export url  to '/export-region' before being decommissioned
- Rename regional export routes to LegacyExportRoutes
- Rename ExportCsvService -> LegacyExportCsvSerice in prep for decommissioning
- Add EmailClient interface for testing email calls
- Add new export database table
- Add export types for new exports: Arrivals & PassengersDaily
- Move gov notify email templates to their own package & classes
- Add new ExportCsvService
- Add new ExportRoutes for new service
- Update and add new tests


ngu04 (approved):
Nice one ! 👍

PR #155, opened 2023-09-13
DRTII-1254 Add route to query status of download. More tests

richbirch:
Add endpoint to query download status
Refactor database use to an interface to make testable


ngu04 (approved):
Nice one !

PR #168, opened 2023-10-26
DRTII-1224 drop-in Notification

ngu04:
- User with more than 2 week of access request and not registered for any drop-ins are send notification to join drop-ins.
- Introduction Mockito to verify logic around email sent 
- Also adding logback-test.xml which helps to debug issue with db when required during test runs


src/main/scala/uk/gov/homeoffice/drt/DrtDashboardApp.scala
```diff
@@ -4,15 +4,15 @@ import akka.actor.typed.ActorSystem
...
+import uk.gov.homeoffice.drt.schedule.{DropInNotification, DropInReminder, UserTracking}
 import uk.gov.service.notify.NotificationClient
 
 import scala.concurrent.duration.DurationInt
 
 object DrtDashboardApp extends App {
   val config = ConfigFactory.load()
 
-  val serverConfig = ServerConfig(
+  lazy val serverConfig = ServerConfig(
```
richbirch: This doesn't need to be lazy now does it?

src/main/scala/uk/gov/homeoffice/drt/db/DropInRegistrationDao.scala
```diff
@@ -63,4 +63,14 @@ case class DropInRegistrationDao(db: Database) {
     result
   }
 
+  def findRegisteredUser(email: String): Future[Seq[DropInRegistrationRow]] = {
```
richbirch: This should probably be called `findRegistrationsByEmail`

src/main/scala/uk/gov/homeoffice/drt/db/UserAccessRequestDao.scala
```diff
@@ -93,4 +99,16 @@ case class UserAccessRequestDao(db: Database) extends IUserAccessRequestDao {
   def selectForStatus(status: String): Future[Seq[UserAccessRequest]] = {
     db.run(userAccessRequests.filter(_.status === status).result)
   }
+
+  def selectApprovedUserAfterSpecificDate(): Future[Seq[UserAccessRequest]] = {
```
richbirch: A couple of things here:

1. We shouldn't need to refer back to this table once a user record is created. It looks like we're missing a `createdAt` field on the user table which mean we don't have to rely on looking in this one. We should add that field and populate it with the `requestTime` value from this table
2. Since this is only used in conjunction with `getUsersWithoutDropInNotification`, we could just use a single query to get the users we're interested in: `usersToSendADropInNotification`. You can join on the other tables to exclude records that already have already been sent a notification and records that have already registered for a drop in

src/main/scala/uk/gov/homeoffice/drt/notifications/EmailNotifications.scala
```diff
@@ -41,6 +42,20 @@ case class EmailNotifications(accessRequestEmails: List[String], client: Notific
     s"https://${curad.getListOfPortOrRegion.headOption.map(_.trim.toLowerCase).getOrElse("")}.$domain/#trainingHub/dropInBooking"
   }
 
+  def getDropInLink(userAccessRequest: UserAccessRequest, domain: String): String = {
```
richbirch: This needs tests around it

src/main/scala/uk/gov/homeoffice/drt/schedule/DropInNotification.scala
```diff
@@ -0,0 +1,66 @@
...
+        dropInService,
+        timers,
+        timerInitialDelay,
+        serverConfig.dropInNotificationFrequency.minutes,
+        serverConfig.rootDomain,
+        context).dropInNotificationReminder)
+    }
+}
+
+class DropInNotification(notifications: EmailNotifications,
```
richbirch, line 37: This needs some tests around it to demonstrate the expected behaviour, eg make sure users aren't receiving multiple emails etc

src/main/scala/uk/gov/homeoffice/drt/services/DropInService.scala
```diff
@@ -18,11 +17,43 @@ class DropInService(dropInDao: DropInDao, dropInRegistrationDao: DropInRegistrat
...
+            usersToNotify.map(user => notifications.sendDropInReminderEmail(user.email, dropIn, teamEmail))
             usersToNotify.map(user => dropInRegistrationDao.updateEmailSentTime(user.dropInId.toString))
           }
       }
     }
   }
 
+
+  def sendDropInNotificationToNewUsers(notifications: EmailNotifications, rootDomain: String)(implicit ec: ExecutionContext) = {
+    val approvedUsers: Future[Seq[UserAccessRequest]] = userRequestService.getApprovedUserAfterSpecificDate()
```
richbirch: This logic would be greatly simplified if we just had a single query to pull out the users we want to send the notifications to, ie a) acive, b) registered over 2 weeks ago, c) hasn't already been sent a notification and d) hasn't already registered for a drop in session.

We'd just loop through the results and update the user record as a notification is sent.

src/test/scala/uk/gov/homeoffice/drt/db/DropInDaoSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+  override protected def before: Any = {
+    Await.ready(
+      db.run(DBIO.seq(
+        TestDatabase.dropInTable.schema.dropIfExists,
+        TestDatabase.dropInTable.schema.createIfNotExists)
+      ), 2.second)
+  }
+
+  "DropInDao" >> {
+    "should return insert and select drop-ins" >> {
```
richbirch: I'm not quite sure what this test is testing. Can it be made a little clearer?

src/test/scala/uk/gov/homeoffice/drt/db/UserAccessRequestDaoSpec.scala
```diff
@@ -0,0 +1,121 @@
...
+        lineManager = "",
+        agreeDeclaration = true,
+        accountType = "port",
+        portOrRegionText = "lhr",
+        staffText = "",
+        status = "Requested",
+        requestTime = new Timestamp(Instant.now().minusSeconds(60).toEpochMilli))), 1.second)
+      val userAccessRequests = Await.result(userAccessRequestDao.selectAll(), 1.second)
+
+      userAccessRequests.size mustEqual 1
```
richbirch: This should probably test the values were saved an returned as expected

src/test/scala/uk/gov/homeoffice/drt/db/UserDaoSpec.scala
```diff
@@ -26,39 +26,45 @@ class UserDaoSpec extends Specification with BeforeEach {
     email = "user1@test.com",
     latest_login = new Timestamp(Instant.now().toEpochMilli),
     inactive_email_sent = None,
-    revoked_access = None)
+    revoked_access = None,
+    drop_in_notification = None
```
richbirch: `drop_in_notification_sent_at`

src/test/scala/uk/gov/homeoffice/drt/services/DropInServiceSpec.scala
```diff
@@ -0,0 +1,294 @@
...
+      val userRequestService = UserRequestService(UserAccessRequestDao(TestDatabase.db))
+      val dropInService: DropInService = new DropInService(dropInDao,
+        dropInRegistrationDao,
+        userService,
+        userRequestService, teamEmail)
+      val emailClient = Mockito.mock(classOf[NotificationClientApi])
+      when(emailClient.sendEmail(any(), any(), any(), any())).thenReturn(new SendEmailResponse(response(templateId = UUID.randomUUID().toString)))
+      val emailNotifications = EmailNotifications(List("test@test.com"), emailClient)
+
+      val accessRequest = getAccessRequest
```
richbirch: If we add the `createdAt` column to the user table then we won't need to look in the access requests anymore

src/test/scala/uk/gov/homeoffice/drt/services/DropInServiceSpec.scala
```diff
@@ -0,0 +1,294 @@
...
+      val userService = UserService(UserDao(TestDatabase.db))
+      val userRequestService = UserRequestService(UserAccessRequestDao(TestDatabase.db))
+      val dropInService: DropInService = new DropInService(dropInDao,
+        dropInRegistrationDao,
+        userService,
+        userRequestService, teamEmail)
+      val emailClient = Mockito.mock(classOf[NotificationClientApi])
+      val emailNotifications = EmailNotifications(List("test@test.com"), emailClient)
+      when(emailClient.sendEmail(any(), any(), any(), any())).thenReturn(new SendEmailResponse(response(templateId = UUID.randomUUID().toString)))
+      val accessRequest = getAccessRequest
```
richbirch: All these setup lines are repeated and can be moved into a setup function. It'll make the test clearer too if you just include the most relevant parts of what you're testing

richbirch (changes requested):
Nice work mate!

There's a good opportunity to simplify some things here, eg adding a created date to the user table so we don't have to rely on the access requests table, and using a single query to pull out all the users we want to send a notification to

richbirch (approved):
Awesomeness

PR #169, opened 2023-10-30
Drtii 1305 alerting on incidents

richbirch:
Add scheduled health check calls for 5 different parameters
When a health check fails 3 consecutive times an alarm is triggered, currently sending an email
When a health check succeeds after entering an alarm state an email is sent to say it is resolved


ngu04 (approved):
Nice one !

ngu04:
Just question here . Once email is triggered for failure above alarm Trigger Consecutive Failures number , does it not trigger another email still it is resolved or keeps triggering email if failure keeps happening ?

richbirch:
> Just question here . Once email is triggered for failure above alarm Trigger Consecutive Failures number , does it not trigger another email still it is resolved or keeps triggering email if failure keeps happening ?

One email is sent when the alarm is triggered, and one email is sent when the alarm is silenced. We might want to add the option for it to repeat the email eg every hour until it's resolved

PR #173, opened 2023-11-14
Drtii 1305 alerting on incidents

richbirch:
Tidy server side routes
Add missing api routes to `ApiClient`
Fix log out menu item link
Set no caching headers for index.html to prevent 'white page of death' (because compiled js files change names on each build)
Upgrade akka & akka http to latest


ngu04 (approved):
Nice one !

PR #178, opened 2023-12-04
Drtii 1316 feedback

ngu04:
- Implementing Feedback form using Formik 
- Api endpoint for feedback 
- Feedback page to view and download csv


database_setup/ab_feature.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE ab_feature
+(
+    email         text NOT NULL,
+    function_name text,
+    presented_at  timestamp NOT NULL,
+    test_type     text,
```
richbirch: `test_type` could be `ab_version`, ie we talk about whether a user is seeing version A or version B of some content

database_setup/user_feedback.sql
```diff
@@ -0,0 +1,15 @@
...
+    actioned_at timestamp NOT NULL,
+    feedback_at timestamp,
+    close_banner boolean,
+    feedback_type text,
+    bf_role     text,
+    drt_quality text,
+    drt_likes   text,
+    drt_improvements text,
+    participation_interest text,
+    a_or_b_test text,
```
richbirch: this is probably also clearer as `ab_version`

react/src/App.tsx
```diff
@@ -109,6 +111,8 @@ export const App = () => {
             <Route path={"list/registered-users/:dropInId"} element={<DropInSessionRegistrations/>}/>
           </Route>
           <Route path={"/health-checks"} element={<HealthCheckEditor/>}/>
+          <Route path={"/feedback/:feedbackType/:aORbTest"} element={<FeedbackForms/>}/>
```
richbirch: maybe `:abVersion`

react/src/components/feedback/FeedbackForms.tsx
```diff
@@ -0,0 +1,404 @@
...
+                What did you think could be improved in DRT ? (optional) :
+              </Typography>
+            </FormLabel>
+            <Typography>If possible, please give examples and provide suggestions</Typography>
+            <Field
+              as={TextField}
+              name="question4"
+              multiline
+              rows={3}
+              value={values.question4}
```
richbirch, line 249: Question 4 displays the answer to question 3 - I assume question 4's value must be incorrectly getting set when question 3 is submitted

react/src/components/feedback/FeedbackForms.tsx
```diff
@@ -0,0 +1,404 @@
...
+            }}>
+              <Field as={RadioGroup} name="question5" value={values.question5} onChange={handleChange}>
+                <FormControlLabel value="Yes" control={<Radio/>} label="Yes"/>
+                <FormControlLabel value="No" control={<Radio/>} label="No"/>
+              </Field>
+            </Stack>
+            <br/>
+            <Button type="submit"
+                    sx={{float: "left", width: 'auto', maxWidth: '120px', padding: '6px 12px'}}
+                    variant="outlined">Continue</Button>
```
richbirch: This final question's button should say `Submit feedback`

react/src/components/feedback/FeedbackList.tsx
```diff
@@ -0,0 +1,135 @@
...
+      link.href = downloadUrl;
+      link.download = filename;
+      document.body.appendChild(link);
+      link.click();
+      document.body.removeChild(link);
+      window.URL.revokeObjectURL(downloadUrl);
+    } catch (error) {
+      console.error('Download failed', error);
+    }
+  };
```
richbirch: I'm not sure we need this. The button can just be a link to the download url that opens in a new tab

react/src/services/ApiClient.ts
```diff
@@ -31,6 +31,9 @@ export default class ApiClient implements IApiClient {
 
   public static exportRegionEndpoint = "/api/export-region"
 
+  public static getFeedBacksEndpoint = "/api/feedback/all";
+  public static saveFeedBacksEndpoint = "/api/feedback/save";
```
richbirch: To be RESTful, there should just be one url `/api/feedback`
The routes should respond according to the request type, ie:
- GET = list of feedback responses
- POST = save new feedback response

react/src/components/feedback/FeedbackList.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  return <Stack gap={4} alignItems={'stretch'} sx={{mt: 2}}>
+    <Breadcrumbs>
+      <Link to={"/"}>
+        Home
+      </Link>
+      <Typography color="text.primary">User Feedbacks</Typography>
```
richbirch: `User feedback responses`

react/src/components/feedback/FeedbackList.tsx
```diff
@@ -0,0 +1,135 @@
...
+    </Breadcrumbs>
+    <Stack direction={'row'} justifyContent={'space-between'}>
+      <Button startIcon={<FileDownloadIcon/>} onClick={handleDownload}>Download Feedbacks</Button>
```
richbirch: `Download feedback responses`

src/main/scala/uk/gov/homeoffice/drt/routes/FeedbackRoutes.scala
```diff
@@ -0,0 +1,100 @@
...
+  implicit val feedbackDataFormatParser: RootJsonFormat[FeedbackData] = jsonFormat7(FeedbackData)
+  implicit val userFeedbackRowFormatParser: RootJsonFormat[UserFeedbackRow] = jsonFormat11(UserFeedbackRow)
+}
+
+object FeedbackRoutes extends FeedbackJsonFormats with BaseRoute {
+
+  val formatter = DateTimeFormatter.ofPattern("dd-MM-yyyy HH:mm:ss")
+  val formattedDate: Timestamp => String = timestamp => timestamp.toLocalDateTime.format(formatter)
+
+  def exportFeedback(feedbackDao: UserFeedbackDao)(implicit ec: ExecutionContext) = path("export") {
```
richbirch: execution context isn't used in this function

src/main/scala/uk/gov/homeoffice/drt/routes/FeedbackRoutes.scala
```diff
@@ -0,0 +1,100 @@
...
+
+trait FeedbackJsonFormats extends DefaultTimeJsonProtocol {
+
+  implicit val feedbackDataFormatParser: RootJsonFormat[FeedbackData] = jsonFormat7(FeedbackData)
+  implicit val userFeedbackRowFormatParser: RootJsonFormat[UserFeedbackRow] = jsonFormat11(UserFeedbackRow)
+}
+
+object FeedbackRoutes extends FeedbackJsonFormats with BaseRoute {
+
+  val formatter = DateTimeFormatter.ofPattern("dd-MM-yyyy HH:mm:ss")
```
richbirch: Can we add the return type to all public functions and values, and set private where appropriate as per intellij's suggestions?

src/main/scala/uk/gov/homeoffice/drt/routes/FeedbackRoutes.scala
```diff
@@ -0,0 +1,100 @@
...
+  val formattedDate: Timestamp => String = timestamp => timestamp.toLocalDateTime.format(formatter)
+
+  def exportFeedback(feedbackDao: UserFeedbackDao)(implicit ec: ExecutionContext) = path("export") {
+    get {
+      val csvHeader: String = "Email,ActionedAt,FeedbackAt,CloseBanner,FeedbackType,BfRole,DrtQuality,DrtLikes,DrtImprovements,ParticipationInterest,AOrBTest"
```
richbirch: What are ActionedAt & FeedbackAt? They appear to contain the same value. Should FeedbackAt be CreatedAt?

The column headings can have spaces in them to make them more readable:
`"Email,Actioned at,Created at,Banner closed,Type,BF Role,Drt Quality,Likes,Improvements,Participation Interest?,AB version"

src/main/scala/uk/gov/homeoffice/drt/routes/FeedbackRoutes.scala
```diff
@@ -0,0 +1,100 @@
...
+            saveFeedbackResult.map(_ => complete(StatusCodes.OK, s"Feedback from user $userEmail is saved successfully")), "Saving feedback")
+        }
+      }
+    }
+  }
+
+
+  def apply(feedbackDao: UserFeedbackDao)(implicit ec: ExecutionContext) =
+    pathPrefix( "feedback") {
+      concat(saveFeedback(feedbackDao) ~ getFeedbacks(feedbackDao) ~ exportFeedback(feedbackDao))
```
richbirch: tilde `~` means concat, so you don't need to wrap in `concat` as well, or you use `concat` and pass each route as comma separated values

src/main/scala/uk/gov/homeoffice/drt/routes/FeedbackRoutes.scala
```diff
@@ -0,0 +1,100 @@
...
+        .concat(fetchDataStream.map(toCsvString).map(str => ByteString(str + "\n")))
+
+      complete(HttpResponse(
+        headers = List(`Content-Disposition`(attachment, Map("filename" -> s"feedback-export-${Instant.now().toEpochMilli}.csv"))),
+        entity = HttpEntity(ContentTypes.`text/csv(UTF-8)`, csvDataStream)))
+
+    }
+  }
+
+  def getFeedbacks(feedbackDao: UserFeedbackDao)(implicit ec: ExecutionContext) = path("all") {
```
richbirch: let's drop the `all` & `save` paths to make it a restful endpoint

src/test/scala/uk/gov/homeoffice/drt/routes/FeedbackRoutesSpec.scala
```diff
@@ -0,0 +1,140 @@
...
+      question_4 = "Staffing",
+      question_5 = "true")
+    val email = "my@email.com"
+
+    Post("/feedback/save", feedbackData.toJson) ~>
+      RawHeader("X-Auth-Roles", BorderForceStaff.name) ~>
+      RawHeader("X-Auth-Email", email) ~> userFeedbackRoute(userFeedbackDao) ~> check {
+      val response = responseAs[String]
+      userFeedbackDao.selectAll().map { userFeedback =>
+        userFeedback.size === 2  && response.contains(s"Feedback from user $email is saved successfully")
```
richbirch: Shouldn't there only be one row returned if we've only saved one set of feedback?

Also, if you drop the `check {` to a new line then intellij will intent the block correctly

react/src/components/feedback/FeedbackForms.tsx
```diff
@@ -0,0 +1,404 @@
...
+
+  const closeFeedback = (
+    <Stack>
+      <Typography variant="h5" sx={{float: "centre", fontWeight: 'bold', color: '#111224'}}>
+        Thank you for your feedback
+      </Typography>
+      <br/>
+      <Button type="submit"
+              sx={{float: "left", width: 'auto', maxWidth: '120px', padding: '6px 12px'}}
+              onClick={() => window.close()}
```
richbirch: `window.close` gives an error `Scripts may close only the windows that were opened by them.`
So we may want to remove the button and just say something like 'You may now close this window'

src/main/scala/uk/gov/homeoffice/drt/routes/FeedbackRoutes.scala
```diff
@@ -0,0 +1,100 @@
...
+  val formatter = DateTimeFormatter.ofPattern("dd-MM-yyyy HH:mm:ss")
+  val formattedDate: Timestamp => String = timestamp => timestamp.toLocalDateTime.format(formatter)
+
+  def exportFeedback(feedbackDao: UserFeedbackDao)(implicit ec: ExecutionContext) = path("export") {
+    get {
+      val csvHeader: String = "Email,ActionedAt,FeedbackAt,CloseBanner,FeedbackType,BfRole,DrtQuality,DrtLikes,DrtImprovements,ParticipationInterest,AOrBTest"
+
+      val fetchDataStream = feedbackDao.selectAllAsStream()
+
+      val toCsvString: UserFeedbackRow => String = feedback => s"${feedback.email}," +
```
richbirch: You should either use a CSV library to create the rows properly, or make sure you wrap all content with quotes so that commas in the answers don't break the CSV (they do at the moment)

richbirch (changes requested):
Nice work!
A few things to fix and tidy up
It would also be worth you reading up on redux. You could easily split the form into separate pages with their own route if the question state was stored using redux. That would be easier to work with than a 400 line component

react/src/components/feedback/FeedbackList.tsx
ngu04: This is because ,I also wanted to  setRequestedAt(moment().valueOf()) when button is click . Didn't find option to update state with Link so went with this one

src/main/scala/uk/gov/homeoffice/drt/routes/FeedbackRoutes.scala
ngu04: Initially I thought keep close banner as separate date but later added flag . So you are correct in identifying we only need one date with closing flag .

src/test/scala/uk/gov/homeoffice/drt/routes/FeedbackRoutesSpec.scala
ngu04: expecting 1 but looks like drop and create table is not deleting rows from previous test

react/src/components/feedback/FeedbackForms.tsx
ngu04: Value is not set in state so it is not reflected in saved  form  .I think is browser which is keeping data in textbox from previous question . let me see if I can explicit make browser to clear it.

react/src/components/feedback/FeedbackList.tsx
richbirch: This works:
```

      <Button
        startIcon={<FileDownloadIcon/>}
        component={MuiLink}
        href={`${ApiClient.feedBacksEndpoint}/export`}
        target={'_blank'}
        onClick={() => setRequestedAt(moment().valueOf())}
      >
        Download feedback responses
      </Button>
      ```

File: src/test/scala/uk/gov/homeoffice/drt/routes/FeedbackRoutesSpec.scala (hunk 15 above)
Comment (richbirch) [reply src/test/scala/uk/gov/homeoffice/drt/routes/FeedbackRoutesSpec.scala]: I think it's worth figuring out why the tables aren't cleared between tests else the tests aren't going to be a true reflection of what's happening

File: src/main/scala/uk/gov/homeoffice/drt/routes/FeedbackRoutes.scala [hunk 18]
```diff
@@ -0,0 +1,102 @@
...
+}
+
+object FeedbackRoutes extends FeedbackJsonFormats with BaseRoute {
+
+  private val formatter = DateTimeFormatter.ofPattern("dd-MM-yyyy HH:mm:ss")
+  private val formattedDate: Timestamp => String = timestamp => timestamp.toLocalDateTime.format(formatter)
+
+  def exportFeedback(feedbackDao: UserFeedbackDao): Route = path("export") {
+    get {
+      val csvHeader: String = "Email ,CreatedAt ,CloseBanner ,FeedbackType ,BfRole ,DrtQuality ,DrtLikes ,DrtImprovements ,ParticipationInterest ,ABVersion"
```
richbirch: I meant add spaces between the words in each column so that they're easier to read in a spreadsheet :)
eg `Email,Created at,Feedback type,BF Role...`

richbirch (changes requested):
Nice one.
Just remaining:
- that download link to fix (a solution in the comments)
- fix the tests not clearing the db table
- csv headers with spaces between words to make easier to read

src/main/scala/uk/gov/homeoffice/drt/routes/FeedbackRoutes.scala
```diff
@@ -32,14 +32,13 @@ object FeedbackRoutes extends FeedbackJsonFormats with BaseRoute {
 
   def exportFeedback(feedbackDao: UserFeedbackDao): Route = path("export") {
     get {
-      val csvHeader: String = "Email ,Created at ,Close banner ,Feedback type ,Bf role ,Drt quality ,Drt likes ,Drt improvements ,Participation interest ,AB version"
+      val csvHeader: String = "Email ,Created at ,Feedback type ,Bf role ,Drt quality ,Drt likes ,Drt improvements ,Participation interest ,AB version"
```
richbirch: Technically there shouldn't be any whitespace either side of a comma

richbirch (approved):
Nice - let's just remove the additional white space in the column headers (the ones before or after a comma)

PR #187, opened 2023-12-12
Drtii 1255 update download manager exported file

richbirch:
Fix email download link (missing /api)
Add port terminals to config
Fix feedback route (user-feedback) so page refresh works

### DIFF
--- a/src/main/scala/uk/gov/homeoffice/drt/routes/IndexRoute.scala
+++ b/src/main/scala/uk/gov/homeoffice/drt/routes/IndexRoute.scala
@@ -29,7 +29,6 @@ case class IndexRoute(urls: Urls, indexResource: Route) {
               RawHeader("Expires", "0"),
             ))(indexRouteDirectives)
           },
-//          getFromResourceDirectory("frontend")
         )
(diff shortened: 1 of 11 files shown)


src/main/scala/uk/gov/homeoffice/drt/routes/IndexRoute.scala
```diff
@@ -38,7 +37,7 @@ case class IndexRoute(urls: Urls, indexResource: Route) {
       pathPrefix("region")(indexRouteDirectives),
       pathPrefix("feature-guides")(indexRouteDirectives),
       pathPrefix("drop-in-sessions")(indexRouteDirectives),
-      pathPrefix("feedback")(indexRouteDirectives),
+      pathPrefix("user-feedback")(indexRouteDirectives),
```
ngu04, line 41: does this also has to match server api ?  'user-feedback' is react route and server route is 'feedback'

ngu04 (commented):
just one comment about route endpoint

ngu04 (approved):
Nice one !

PR #188, opened 2023-12-13
DRT-1251 Download Manager

jacksonhyde:
- Implements download manager route and components
- Implements redux and redux-saga


react/src/components/Navigation.tsx
```diff
@@ -30,6 +30,7 @@ export default function Navigation(props: IProps) {
     {label: 'Access requests', link: '/access-requests', roles: ['manage-users']},
     {label: 'Alert notices', link: '/alerts', roles: ['manage-users']},
     {label: 'Drop-in sessions', link: '/drop-in-sessions', roles: ['manage-users']},
+    {label: 'Download Manager', link: '/download', roles: ['manage-users']},
```
richbirch: I'm not actually sure which users we're releasing this to. The `manage-users` role means it'll only be available to our own team. We should probably add the RCC roles, which are:
```
rcc:central
rcc:heathrow
rcc:north
rcc:south
```

react/src/components/downloadmanager/DownloadManager.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+  const isRccRegion = (regionName : string) => {
+    return user.roles.includes("rcc:" + regionName.toLowerCase())
+  }
+
+  let interval: {current: ReturnType<typeof setInterval> | null | any} = React.useRef(null);
+
+  React.useCallback(() => {
+    
+  }, [checkDownloadStatus]);
```
richbirch: Do we need this callback if it's not doing anything?

react/src/components/downloadmanager/DownloadManager.tsx
```diff
@@ -0,0 +1,212 @@
...
+    end: moment(),
+  });
+  const [breakdown, setBreakdown] = React.useState<string>('passengers-port');
+  const [daily, setDaily] = React.useState<boolean>(false);
+
+  const isRccRegion = (regionName : string) => {
+    return user.roles.includes("rcc:" + regionName.toLowerCase())
+  }
+
+  let interval: {current: ReturnType<typeof setInterval> | null | any} = React.useRef(null);
```
richbirch, line 47: I think the correct type would be `NodeJS.Timeout | null` as `setInterval` returns a `NodeJS.Timeout`

react/src/components/downloadmanager/DownloadManager.tsx
```diff
@@ -0,0 +1,212 @@
...
+    setModalOpen(false)
+    clearInterval(interval.current);
+  }
+
+  const onBreakdownChange = (event: React.ChangeEvent<HTMLInputElement>, breakdown: string) => {
+    setBreakdown(breakdown);
+  }
+
+  const handlePortChange = (event: SelectChangeEvent<typeof selectedPorts>) => {
+    const {
```
richbirch, line 83: This is fun! I'd just have written it `const value = event.target.value`

react/src/components/downloadmanager/DownloadManager.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+  const handlePortCheckboxGroupChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const {
+      target: { name, checked },
+    } = event;
+    let region = userPortsByRegion.filter(region => region.name === name)[0]
+    if (checked) {
+      // add all in region
+      let newSelection = [...selectedPorts, ...region.ports]
+      let deduped = newSelection.filter((element, index) => {
```
richbirch: These `let`s could all be `const` as far as I can tell - preferring immutability etc

react/src/components/downloadmanager/DownloadManager.tsx
```diff
@@ -0,0 +1,212 @@
...
+        return newSelection.indexOf(element) === index;
+      });
+      setSelectedPorts(deduped)
+    } else {
+      // remove any selected from region
+      setSelectedPorts(selectedPorts.filter(port => !region.ports.includes(port)))
+    }
+  }
+
+  const disablePassengerBreakdown = (): boolean =>  {
```
richbirch, line 114: This is cool  - I actually didn't know how to write the return type on a parameterless function. Thanks!

react/src/components/downloadmanager/DownloadPorts.tsx
```diff
@@ -0,0 +1,130 @@
...
+                  <FormControlLabel control={<Checkbox name={region.name} onChange={handlePortCheckboxGroupChange} inputProps={{ 'aria-label': 'controlled' }} />} label={region.name} />
+                  <Chip sx={{marginLeft: '1em'}} label={`${region.ports.filter(port => selectedPorts.includes(port)).length} selected`} />
+
+              </AccordionSummary>
+              <AccordionDetails>
+                <Grid container spacing={4}>
+                  <Grid item>
+                    <List sx={{paddingRight: 1, borderRight: region.ports.length > 4 ? '1px solid #eee': 'none'}}>
+                      { region.ports.slice(0,4).map((airport: string) => {
+                          return <ListItem key={airport}><FormControlLabel key={airport} control={<Checkbox name={airport} checked={selectedPorts.includes(airport)} onChange={handlePortCheckboxChange} inputProps={{ 'aria-label': 'controlled' }} />} label={airport} /></ListItem>
```
richbirch: There's scope to move the map contents into a function to reduce the repetition across the 4 rows. No biggie - up to you..

react/src/components/downloadmanager/downloadManagerSagas.ts
```diff
@@ -0,0 +1,90 @@
...
+    startDate,
+    endDate
+  };
+};
+
+function* handleRequestDownload(action: RequestDownloadActionType) {
+  try {
+    yield put(setStatus('requested'))
+
+    let payload : RequestDownloadPayload = {
```
richbirch: Nice - the saga magic :)
I think all the `let`s here can also be `const`s

react/src/services/ApiClient.ts
```diff
@@ -30,6 +30,8 @@ export default class ApiClient implements IApiClient {
   public static featureGuidesUpdatePublishedEndpoint = "/api/feature-guides/update-published"
 
   public static exportRegionEndpoint = "/api/export-region"
+  public static exportEndpoint = "/api/export"
+  public static exportStatusEndpoint = "/api/export"
```
richbirch: Guess we only need one of these as they're the same value?

richbirch (approved):
Nice work mate. I just had a few suggestions for tidying, and we'll need to add more roles to the navigation item so non-super users can see it, but otherwise looks good to go!

react/src/components/Navigation.tsx
jacksonhyde: Ah hah that makes sense!

react/src/components/downloadmanager/DownloadManager.tsx
jacksonhyde: No definitely don't need - I thought we might need to use it for the status timeout but it wasn't needed in the end, must have forgotten to remove

react/src/components/downloadmanager/DownloadManager.tsx
jacksonhyde: That's what I thought, but the linter throws errors on the clearInterval call, after much faffing about this was the only thing that seemed to build.

richbirch: bloody linters hey!

react/src/components/downloadmanager/DownloadManager.tsx
jacksonhyde: It is fun isn't it! It's good when you need to declare multiple vars from events (line 102 is a good example). Admittedly when you're just grabbing the value it feels a little OTT but I thought I'd keep it consistent across the handler actions

react/src/components/downloadmanager/DownloadManager.tsx
jacksonhyde: ah yes ofc, not sure how I let that happen ;)

richbirch: very drole!

react/src/components/downloadmanager/DownloadPorts.tsx
jacksonhyde: Good shout! Should dry things out a bit

react/src/services/ApiClient.ts
jacksonhyde: yes, not quite sure why I doubled up!

react/src/components/downloadmanager/DownloadPorts.tsx
richbirch: pun city :rofl:

richbirch (approved):
Smashing!

PR #189, opened 2023-12-13
Drtii 1336 access requests details box display

richbirch:
Fix layout issue with access requests:
 - add max height & scrollbar
 - format long text responses with line breaks
 - rename a few field labels

Fix/amend a few download manager things:
 - send download link url in response
 - fix download status url
 - stop progress spinner when status != 'preparing'
 - open download in new tab


jacksonhyde (approved):
Good stuff! Nice work figuring out how to type those yields, I didn't realise it was that easy!

PR #195, opened 2023-12-14
DRTII-1316 Rob review comments

ngu04:
- Review comments from Rob mentioned in origin DRTII-1316

### DIFF
--- a/react/src/components/feedback/FeedbackForms.tsx
+++ b/react/src/components/feedback/FeedbackForms.tsx
@@ -102,7 +102,7 @@ export function FeedbackForms() {
               </Field>
             </Stack>
             <br/>
-            <Button type="submit" variant="outlined">Continue</Button>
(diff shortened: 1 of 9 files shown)


react/src/components/feedback/FeedbackForms.tsx
```diff
@@ -347,7 +372,14 @@ export function FeedbackForms() {
...
             <Button type="submit"
-                    sx={{float: "left", width: 'auto', maxWidth: '180px', padding: '6px 12px'}}
+                    sx={{
+                      textTransform: 'none',
+                      float: "left",
+                      width: 'auto',
+                      maxWidth: '200px',
+                      padding: '6px 12px',
+                      fontSize: drtTheme.typography.button.fontSize,
+                    }}
```
richbirch: Most if not all these styles can be removed. Might just need to keep the maxWidth or wrap the button in a container with a max width

ngu04: Correct .  I will do that now !

PR #196, opened 2023-12-14
Drt 1359 download manager validations

jacksonhyde:
- Introduces Validation service
- Validates Download Manager form answers
- Adds validator dependency


richbirch (approved):
Looks good mate

Have you ever used Formik? Combined with Yup it makes form validation really easy to handle as you just define the validation rules which can even be conditional on other fields: https://formik.org/docs/guides/validation

PR #199, opened 2024-01-03
Port Select Delete UI

jacksonhyde:
- Adds a delete button to the Port chips in the download manager port select


ngu04 (approved):
Nice One !

PR #200, opened 2024-01-05
Drtii 1318 port config export

ngu04:
- Port config export

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,7 +19,7 @@ lazy val slickVersion = "3.4.1"
 lazy val awsJava2SdkVersion = "2.16.104"
 lazy val postgresqlVersion = "42.7.0"
 lazy val mockitoVersion = "4.6.1"
-
+lazy val poiScalaVersion ="0.23"
 lazy val root = (project in file(".")).
   settings(
     inThisBuild(List(
@@ -51,7 +51,7 @@ lazy val root = (project in file(".")).
       "com.github.tototoshi" %% "scala-csv" % scalaCsvVersion,
       "org.scalactic" %% "scalactic" % scalaTestVersion,
       "software.amazon.awssdk" % "s3" % awsJava2SdkVersion,
-
+      "info.folone" %% "poi-scala" % poiScalaVersion,
       "com.typesafe.slick" %% "slick" % slickVersion,
       "com.typesafe.slick" %% "slick-hikaricp" % slickVersion,
       "org.postgresql" % "postgresql" % postgresqlVersion,
--- a/build.sbt
+++ b/build.sbt
@@ -19,7 +19,7 @@ lazy val slickVersion = "3.4.1"
 lazy val awsJava2SdkVersion = "2.16.104"
 lazy val postgresqlVersion = "42.7.0"
 lazy val mockitoVersion = "4.6.1"
-lazy val poiScalaVersion ="0.23"
+lazy val poiScalaVersion ="0.24"
 lazy val root = (project in file(".")).
   settings(
     inThisBuild(List(
--- a/react/src/components/ExportConfig.tsx
+++ b/react/src/components/ExportConfig.tsx
@@ -2,7 +2,6 @@ import React from "react";
 import Button from "@mui/material/Button";
 import FileDownloadIcon from "@mui/icons-material/FileDownload";
 import ApiClient from "../services/ApiClient";
-import moment from "moment-timezone";
 
 export function ExportConfig() {
   return <div>
--- a/react/src/components/ExportConfig.tsx
+++ b/react/src/components/ExportConfig.tsx
@@ -1,0 +1,18 @@
+import React from "react";
+import Button from "@mui/material/Button";
+import FileDownloadIcon from "@mui/icons-material/FileDownload";
+import ApiClient from "../services/ApiClient";
+import moment from "moment-timezone";
+
+export function ExportConfig() {
+  return <div>
+    <h1>Export Config</h1>
+    <Button
+      sx={{maxWidth: '350px'}}
+      startIcon={<FileDownloadIcon/>}
+      href={`${ApiClient.exportConfigEndpoint}`}
+      target="_blank"
+      onClick={() => console.log(moment().valueOf())}
+    > Download ports config</Button>
+  </div>
+}
--- a/react/src/components/ExportConfig.tsx
+++ b/react/src/components/ExportConfig.tsx
@@ -12,7 +12,6 @@ export function ExportConfig() {
       startIcon={<FileDownloadIcon/>}
       href={`${ApiClient.exportConfigEndpoint}`}
       target="_blank"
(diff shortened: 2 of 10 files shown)


build.sbt
```diff
@@ -19,7 +19,7 @@ lazy val slickVersion = "3.4.1"
 lazy val awsJava2SdkVersion = "2.16.104"
 lazy val postgresqlVersion = "42.7.0"
 lazy val mockitoVersion = "4.6.1"
-
+lazy val poiScalaVersion ="0.23"
```
richbirch: Looks like version 0.24 was released a couple of weeks ago

react/src/components/ExportConfig.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export function ExportConfig() {
+  return <div>
+    <h1>Export Config</h1>
+    <Button
+      sx={{maxWidth: '350px'}}
+      startIcon={<FileDownloadIcon/>}
+      href={`${ApiClient.exportConfigEndpoint}`}
+      target="_blank"
+      onClick={() => console.log(moment().valueOf())}
```
richbirch: log to be removed

src/main/scala/uk/gov/homeoffice/drt/routes/ExportConfigRoutes.scala
```diff
@@ -0,0 +1,83 @@
...
+
+      val processData = Source(endpoints).mapAsync(10) { case (portCode, uriString) =>
+        val request = HttpRequest(uri = Uri(uriString))
+        httpClient.send(request).flatMap { response =>
+          Unmarshal(response.entity).to[String].map { data =>
+            val lines = data.split("\n")
+            (portCode, lines.toSeq)
+          }
+        }.recover { case e: Throwable =>
+          (portCode, Seq(s"Error while requesting export for $uriString, ${e.getMessage}"))
```
richbirch: We probably don't want to put the uri and the error message in the csv. Maybe say `We couldn't fetch the config for this port. Please try exporting the file again`

src/main/scala/uk/gov/homeoffice/drt/routes/ExportConfigRoutes.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  def getMergePortConfig(httpClient: HttpClient, enabledPorts: Seq[PortCode])(implicit ec: ExecutionContext, mat: Materializer): Route = get {
+    implicit val system: ActorSystem = mat.system
+
+    val endpoints = enabledPorts.map { portCode =>
+      (portCode.iata, s"http://${portCode.iata.toLowerCase}:9000/export/port-config")
+    }
+
+
+    def generateExcelContent(): Source[ByteString, _] = {
```
richbirch: This doesn't need to be a function since you're immediately calling it on the line below.

Instead you can remove the `def generateEx...` and just have `val excelSource = Source.future(processData.map { _ =>`

src/test/scala/uk/gov/homeoffice/drt/routes/ExportConfigRoutesSpec.scala
```diff
@@ -0,0 +1,238 @@
...
+import akka.http.scaladsl.model.{ContentType, MediaTypes}
+import akka.http.scaladsl.testkit.ScalatestRouteTest
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+import uk.gov.homeoffice.drt.MockHttpClient
+import uk.gov.homeoffice.drt.ports.PortCode
+import org.apache.poi.xssf.usermodel.XSSFWorkbook
+import java.io.ByteArrayInputStream
+
+class ExportConfigRoutesSpec extends AnyWordSpec with Matchers with ScalatestRouteTest {
```
richbirch, line 12: Oops the indentation has gone to 4 spaces again for some reason!

src/main/scala/uk/gov/homeoffice/drt/routes/ExportConfigRoutes.scala
```diff
@@ -0,0 +1,83 @@
...
+      val workbook = new XSSFWorkbook()
+
+      def writeToWorkbook(sheetName: String, data: Seq[String]): Unit = {
+        val sheet = workbook.createSheet(sheetName)
+        data.zipWithIndex.foreach { case (value, index) =>
+          val row = sheet.createRow(index)
+          value.split(",").zipWithIndex.foreach { case (cellValue, cellIndex) =>
```
richbirch: This won't correctly deal with cells that contain a comma. We need to handle csv parsing properly. I remember pointing this out in another PR recently - did you use a csv library to parse that time?

src/test/scala/uk/gov/homeoffice/drt/routes/ExportConfigRoutesSpec.scala
```diff
@@ -0,0 +1,238 @@
...
+
+    def mockHttpClient(csvContent: String): MockHttpClient = MockHttpClient(() => csvContent)
+
+    "Request test port config" should {
+        "e-gate schedule data" in {
```
richbirch: It looks like these 6 tests cases are all testing the same thing. I don't think there's any need for more than one, and it should test for csv content with commas in the cells

richbirch (changes requested):
Nice one mate

Just one gotcha that we need to fix - parsing CSVs with commas in the cells. We might not have any commas right now, but if that changes in the future then this export will break.

Other than that there's just a little bit of tidying :+1:

build.sbt
ngu04: I was not able to download 0.24 version

src/test/scala/uk/gov/homeoffice/drt/routes/ExportConfigRoutesSpec.scala
ngu04: Need to see why this happens , May I can get export of style from your intellij

src/main/scala/uk/gov/homeoffice/drt/routes/ExportConfigRoutes.scala
ngu04: Good point , Will add that .

src/test/scala/uk/gov/homeoffice/drt/routes/ExportConfigRoutesSpec.scala
ngu04: Thought to test all data text and it seems fine .

richbirch: The point is that each test is just testing that it splits a line into cells. One test is good enough to test that as the content doesn't matter as long as it contains more than one cell to prove the behaviour works

richbirch (approved):
Nice! I think we're just missing a test for parsing a csv cell containing commas

PR #203, opened 2024-01-12
Drtii 1339 healthcheck slack

ngu04:
Drtii 1339 : Adding slack notification for health check

### DIFF
--- a/src/main/scala/uk/gov/homeoffice/drt/Server.scala
+++ b/src/main/scala/uk/gov/homeoffice/drt/Server.scala
@@ -190,17 +190,16 @@ object Server {
     implicit val timeout: Timeout = new Timeout(1.second)
     implicit val scheduler: Scheduler = system.scheduler
 
-    def sendSlackNotification(portCode: PortCode, checkName: String, priority: IncidentPriority): Unit = {
+    def sendSlackNotification(portCode: PortCode, checkName: String, priority: IncidentPriority, status: String): Unit = {
       val port = portCode.toString.toUpperCase
       val link = urls.urlForPort(port)
-      val message =
-       s"""
-          Health Check Alert
-          port: $port
-          name: $checkName
-          priority: ${priority.toString}
-          link: $link
-        """
+      val message = {
+        s"""Health Check Alert - $status
+           |port: $port
+           |name: $checkName
+           |priority: ${priority.toString}
+           |link: $link""".stripMargin
+      }
       slackClient.notify(message)
     }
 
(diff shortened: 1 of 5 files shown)


src/main/scala/uk/gov/homeoffice/drt/Server.scala
```diff
@@ -187,6 +190,20 @@ object Server {
...
+      val port = portCode.toString.toUpperCase
+      val link = urls.urlForPort(port)
+      val message =
+       s"""
+          Health Check Alert
+          port: $port
+          name: $checkName
+          priority: ${priority.toString}
+          link: $link
+        """
```
richbirch: Is there a reason for leaving blank lines, and starting every line with a 10 space indent?

src/main/scala/uk/gov/homeoffice/drt/notifications/SlackClient.scala
```diff
@@ -0,0 +1,30 @@
...
+import org.slf4j.LoggerFactory
+import uk.gov.homeoffice.drt.HttpClient
+
+import scala.concurrent.ExecutionContext
+
+trait SlackClient {
+  def notify(message: String)(implicit ec: ExecutionContext, mat: Materializer): Unit
+}
+
+case class SlackClientImpl(httpClient: HttpClient, webhookUrl: String) extends SlackClient {
```
richbirch, line 14: How do we specify the channel messages are sent to?

richbirch, line 14: Can we add tests for this new client?

src/main/scala/uk/gov/homeoffice/drt/Server.scala
```diff
@@ -196,10 +213,15 @@ object Server {
...
+    val soundAlarm = (portCode: PortCode, checkName: String, priority: IncidentPriority) => {
       sendEmail(portCode, checkName, priority, serverConfig.healthCheckTriggeredNotifyTemplateId)
-    val silenceAlarm = (portCode: PortCode, checkName: String, priority: IncidentPriority) =>
+      sendSlackNotification(portCode, checkName, priority)
+    }
+
+    val silenceAlarm = (portCode: PortCode, checkName: String, priority: IncidentPriority) => {
       sendEmail(portCode, checkName, priority, serverConfig.healthCheckResolvedNotifyTemplateId)
+      sendSlackNotification(portCode, checkName, priority)
+    }
```
richbirch: The messages don't say if the alarm is triggered or resolved

richbirch (changes requested):
Nice one.
A few things to address

src/main/scala/uk/gov/homeoffice/drt/Server.scala
ngu04: Just had few message in dev channel and they look okay with 10 space . But I think it we will be fine without spaces as well .

src/main/scala/uk/gov/homeoffice/drt/notifications/SlackClient.scala
ngu04: There only one webhook url which is specific to channel for a env (it is retrieved from secret) . Preprod would have alerts-preprod and prod would have alerts-prod .

ngu04: sure

src/main/scala/uk/gov/homeoffice/drt/Server.scala
ngu04: That is good point .

richbirch (approved):
Nice one mate

PR #211, opened 2024-01-23
DRT-1391, DRT-1391, DRT-1395 - Download manager fixes

jacksonhyde:
- Adds close button to download manager modal (DRT-1393)
- Better handling of network errors (DRTII-1391)
- Improves accordion expanding logic (DRTII-1395)

### DIFF
--- a/react/src/components/downloadmanager/DownloadPorts.tsx
+++ b/react/src/components/downloadmanager/DownloadPorts.tsx
@@ -126,9 +126,6 @@ export default function DownloadPorts({error, handlePortChange, handleRemovePort
               key={region.name} 
               expanded={expandedTab === region.name} 
               onChange={(event, isClosed) => accordionChange(event, isClosed, region.name, region.ports)}
-              sx={{
-
-              }}
               >
               <AccordionSummary
                 expandIcon={<ExpandMoreIcon />}
--- a/react/src/components/downloadmanager/DownloadPorts.tsx
+++ b/react/src/components/downloadmanager/DownloadPorts.tsx
@@ -41,6 +41,7 @@ function getStyles(port: string, selectedPorts: readonly string[], theme: Theme)
 
 export default function DownloadPorts({error, handlePortChange, handleRemovePort, handlePortCheckboxChange, handlePortCheckboxGroupChange, portsByRegion, selectedPorts  }: DownloadPortsProps) {
   const theme = useTheme();
+  const [expandedTab, setExpandedTab] = React.useState<string>('');
 
   const allUserPorts :string[] = portsByRegion.map((region) => [...region.ports]).flat();
   portsByRegion.forEach((region) => region.ports.sort());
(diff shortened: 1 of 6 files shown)


react/src/components/downloadmanager/DownloadPorts.tsx
```diff
@@ -63,6 +87,7 @@ export default function DownloadPorts({error, handlePortChange, handleRemovePort
             multiple
             value={selectedPorts}
             onChange={handlePortChange}
+            sx={{fontSize: '22px'}}
```
richbirch, line 90: If `22px` matches what Franco has in the DRT design specifications then we could add it to the `drtTheme` for a `Select` component

react/src/components/downloadmanager/DownloadPorts.tsx
```diff
@@ -93,16 +118,25 @@ export default function DownloadPorts({error, handlePortChange, handleRemovePort
...
           portsByRegion.map((region: Region) => {
             return (
-            <Accordion key={region.name}>
+            <Accordion 
+              key={region.name} 
+              expanded={expandedTab === region.name} 
+              onChange={(event, isClosed) => accordionChange(event, isClosed, region.name, region.ports)}
+              sx={{
+
+              }}
```
richbirch: Can remove this empty `sx`

react/src/components/downloadmanager/downloadManagerSagas.ts
```diff
@@ -66,13 +66,15 @@ function* handleRequestDownload(action: RequestDownloadActionType) {
       }
     }
 
-    const response: Response = yield call (axios.post, ApiClient.exportEndpoint, payload)
+    const response: Response = yield call (axios.post, ApiClient.exportEndpoint, payload);
+    console.log(typeof response);
```
richbirch: log to remove

richbirch (approved):
Nice one. Just 3 minor suggestions / tidying

react/src/components/downloadmanager/DownloadPorts.tsx
jacksonhyde: This doesn't actually translate down to any actual text, the purpose is to vertically space the select box content so that it matches the height of the Chips when no Chips are present. Otherwise the element grows in height a little when you select something, pushing the whole form down.

This select only ever contains Chips, which have their own text size set by the system.

richbirch: aha gotcha

react/src/components/downloadmanager/DownloadPorts.tsx
jacksonhyde: Resisting the urge to make lewd puns

react/src/components/downloadmanager/downloadManagerSagas.ts
jacksonhyde: oops thought I got them all!

PR #213, opened 2024-01-29
DRTII-1405 Passengers route to proxy call to back end

richbirch:
Add route to proxy passenger number requests to the port app back ends


ngu04 (approved):
Nice one !

PR #215, opened 2024-02-13
Drtii 1411 adjust pax summaries end point

richbirch:
Accept comma separated list of port codes
Collect multiple port responses & respond with json or text as requested
Update tests for multi-port support


ngu04 (approved):
Nice one !

PR #223, opened 2024-02-23
Drtii 1414 king crown

ngu04:
- replacing Logo for GOV.UK
- Also fixing favicon location


react/public/images/gov.uk-black.svg
richbirch, line 1: Do we need this version of the logo? I can only see the white version referenced in App.tsx

richbirch (approved):
Nice one. Just wondering if we need the black version as I can't see it referenced anywhere

PR #234, opened 2024-03-01
Drtii 1427 bst timestamp

ngu04:
- converting UTC timestamp to BST before sending as API data

### DIFF
--- a/src/main/scala/uk/gov/homeoffice/drt/routes/DropInSessionsRoute.scala
+++ b/src/main/scala/uk/gov/homeoffice/drt/routes/DropInSessionsRoute.scala
@@ -7,54 +7,40 @@ import akka.http.scaladsl.server.Route
 import org.joda.time.DateTime
 import org.slf4j.{Logger, LoggerFactory}
 import spray.json.{RootJsonFormat, enrichAny}
-import uk.gov.homeoffice.drt.db.{DropInDao, DropInRow}
+import uk.gov.homeoffice.drt.db.{DropIn, DropInDao, DropInRow}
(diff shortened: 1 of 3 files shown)


richbirch (changes requested):
There's a cleaner way to handle the time zone issue. If you send the unix timestamp from the front end then there's nothing to convert.

You can call `valueOf()` on a moment to get the milliseconds. Then on the server side you just construct a timestamp using this.
Send the timestamp milliseconds to the front end, and moment will automatically use the browser's locale settings to display the correct timezone.
Bob's your uncle, there's no timezone to worry about

src/main/scala/uk/gov/homeoffice/drt/routes/DropInSessionsRoute.scala
```diff
@@ -7,44 +7,42 @@ import akka.http.scaladsl.server.Route
...
 
 case class DropInPublished(published: Boolean)
 
-case class DropInData(title: String, startTime: String, endTime: String, meetingLink: String)
+case class DropInData(title: String, startTime: Long, endTime: Long, meetingLink: String)
 
 trait DropInJsonFormats extends DefaultTimeJsonProtocol {
 
   implicit val dropInDataFormatParser: RootJsonFormat[DropInData] = jsonFormat4(DropInData)
   implicit val dropInRowFormatParser: RootJsonFormat[DropInRow] = jsonFormat7(DropInRow)
```
richbirch: I think this json format for the row class is no longer needed

richbirch (approved):
Nice! I think we can remove the json format for the row class now

PR #239, opened 2024-03-12
Regional dashboards 3

jacksonhyde:
- Refactors dashboard to better split up API calls


react/src/components/regionalpressure/regionalPressureSagas.ts
```diff
@@ -59,57 +59,61 @@ type Response = {
...
     yield(put(setStatus('loading')))
-    const start = moment(action.startDate).startOf('day').format('YYYY-MM-DD');
-    const end = action.searchType === 'single' ? start : moment(action.endDate).startOf('day').format('YYYY-MM-DD');
+    let start = moment(action.startDate);
+    let end = action.searchType === 'single' ? start : moment(action.endDate).endOf('day');
     const historicStart = moment(start).subtract(1, 'year').format('YYYY-MM-DD')
     const historicEnd = moment(end).subtract(1, 'year').format('YYYY-MM-DD')
-    const interval = action.searchType === 'single' ? 'hourly' : 'daily';
+    const duration = moment.duration(end.diff(start)).asHours();
+    let interval = duration >= 48 ? 'daily' : 'hourly';
```
richbirch: I think `start`, `end` & `interval` can each be `const`s

richbirch (approved):
Nice one

react/src/components/regionalpressure/regionalPressureSagas.ts
jacksonhyde: ah yes good spot!

PR #243, opened 2024-03-15
Drtii 1438

ngu04:
- Supporting image in feature guide

### DIFF
--- a/react/src/services/ApiClient.ts
+++ b/react/src/services/ApiClient.ts
@@ -27,7 +27,7 @@ export default class ApiClient implements IApiClient {
   public static dropInSessionRegistrationDeleteEndpoint = "/api/drop-in-register"
 
   public static featureGuidesEndpoint = "/api/feature-guides"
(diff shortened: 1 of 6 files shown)


react/src/services/ApiClient.ts
```diff
@@ -27,7 +27,7 @@ export default class ApiClient implements IApiClient {
   public static dropInSessionRegistrationDeleteEndpoint = "/api/drop-in-register"
 
   public static featureGuidesEndpoint = "/api/feature-guides"
-  public static featureGuidesVideosEndpoint = "/api/feature-guides/get-feature-videos"
+  public static featureGuidesFilesEndpoint = "/api/feature-guides/get-feature-files"
```
richbirch: I think this should be singular, and probably drop the 'get' and a GET request would imply getting results. So `/api/feature-guides/feature-file`

richbirch (approved):
Just to change that end point name, but otherwise good to go :)

PR #253, opened 2024-04-04
Fixing chart exceed calculation

jacksonhyde:
- Simplifying (and fixing exceed count)

### DIFF
--- a/react/src/components/regionalpressure/RegionalPressureChart.tsx
+++ b/react/src/components/regionalpressure/RegionalPressureChart.tsx
@@ -48,8 +48,8 @@ interface RegionalPressureChartProps {
   };
 }
 
-const doesExceed = (forecast: number, historic: number): boolean => {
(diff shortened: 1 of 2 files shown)


react/src/components/regionalpressure/RegionalPressureChart.tsx
```diff
@@ -48,8 +48,8 @@ interface RegionalPressureChartProps {
   };
 }
 
-const doesExceed = (forecast: number, historic: number): boolean => {
-  return forecast > historic;
+const doesExceed = (forecast: number): boolean => {
```
richbirch, line 51: It would be nice to put utility functions like this into another file and write some unit tests around them to check for expected behaviour and protect against accidental future regression

richbirch (approved):
Nice one. As commented, would be great to start unit testing some of the functions in use

PR #254, opened 2024-04-08
Download Manager same date

jacksonhyde:
- Allows same start and end date in the download manager
- Defaults to today rather than yesterday


richbirch (approved):
:poodle:

PR #268, opened 2024-05-14
Regional dashboard chart fix 3

jacksonhyde:
- Tweaks and formatting


react/src/components/regionalpressure/RegionalDashboardDetail.tsx
```diff
@@ -125,13 +131,14 @@ const RegionalPressureDetail = ({ config, portData, historicPortData, interval,
           </Stack>
         </Grid>
         {regionPorts && regionPorts.map((port: string) => {
+          let portName = port.replace("-", ' ')
```
richbirch: this can be a `const`

richbirch (approved):
Nice one!

react/src/components/regionalpressure/RegionalDashboardDetail.tsx
jacksonhyde: Good spot!

PR #271, opened 2024-05-22
Drtii 1082 oauth proxy

ngu04:
- Changing header used for Oauth Proxy which are different to Keycloak gatekeeper


richbirch (approved):
:foot:

PR #278, opened 2024-06-25
Role splits with semicolon to fix

ngu04:
DRTII-1082 : Fixing the role splits in dashboard with oauth-proxy changes


richbirch (approved):
Nice. Does this fix Nick's issue?

ngu04:
This is not fix Nicks issue . There is other issue with National Dasshboard not visible

PR #279, opened 2024-06-26
DRTII-1516 replace reduce with fold to handle empty responses. Add test

richbirch:
Use fold to protect against empty streams from empty responses from ports with no forecast flights


ngu04 (approved):
👍

PR #283, opened 2024-07-08
Drtii 1521 regional dashboard slow date ranges

richbirch:
Directly query the database for passenger summaries, rather that calling port app end points and aggregating into one


ngu04 (approved):
Nice one !

PR #306, opened 2024-11-27
Drtii 1574 accessibility

ngu04:
Adding Footer and accessibility statement


react/src/App.tsx
```diff
@@ -85,63 +85,88 @@ export const App = () => {
     }
   }
 
+  const onClickAccessibilityStatement = () => {
+    navigate('/accessibility/statement');
+  }
+
+  const getRandomABTest = (): string => {
```
richbirch: Remove `Test` from name, right?

richbirch (approved):
Nice one! Just to remove 'Test' from the AB function name

react/src/App.tsx
ngu04: It is for AB test so may be keeping Test would it better ?

PR #309, opened 2024-12-05
Drtii 1696 change api response to flat structure

richbirch:
Flatten API response JSON to remove port & terminal levels
Add port and terminal into flights & queue-slot periods


ngu04 (approved):
👍

PR #311, opened 2024-12-16
Drtii 1686 bx data upload

richbirch:
Add border crossing
 - Import service
 - End point to receive excel file
 - Admin page for uploading
 - tests


src/main/scala/uk/gov/homeoffice/drt/routes/BorderCrossingRoutes.scala
```diff
@@ -0,0 +1,52 @@
...
+import scala.concurrent.Future
+import scala.concurrent.duration.DurationInt
+import scala.util.{Failure, Success}
+
+
+object BorderCrossingRoutes {
+  private val log = LoggerFactory.getLogger(getClass)
+
+  private def tempDestination(fileInfo: FileInfo): File =
+    Files.createTempFile(fileInfo.fileName, ".tmp").toFile
```
ngu04, line 28: is this file get deleted automatically from .tmp ?

ngu04 (approved):
Nice one ! Just has doubt if file created in .tmp deleted

src/main/scala/uk/gov/homeoffice/drt/routes/BorderCrossingRoutes.scala
richbirch: Good point!

PR #334, opened 2025-02-13
DRTII-1725 Retain dates between national and regional views

richbirch:
Fix end comparison end date based on forecast end date
Fix retention of chosen dates when switching between national and regional views


ngu04 (approved):
Nice one !

PR #345, opened 2025-03-12
Drtii 1687 bx data on national and regional dashboards

richbirch:
Switch national & regional dashboards to use BX data for historic comparisons


src/main/scala/uk/gov/homeoffice/drt/routes/PassengerRoutes.scala
```diff
@@ -22,7 +22,7 @@ import scala.util.{Failure, Success}
 object PassengerRoutes {
   private val log = org.slf4j.LoggerFactory.getLogger(getClass)
 
-  def apply(summaryProvider: (LocalDate, LocalDate, Granularity, Option[Terminal]) => PortCode => Source[(Map[Queue, Int], Int, Option[Any]), NotUsed])
```
ngu04, line 25: Wondering if we create case class SummaryProvider(LocalDate, LocalDate, Granularity, Option[Terminal]) 
and case class Queues(Map[Queue, Int], Int, Map[Queue, Int], Option[Any]) readability of code could be better ?

react/src/components/regionalpressure/regionalPressureSagas.ts
```diff
@@ -85,24 +85,72 @@ export const requestPaxTotals = (
...
+  const egatePax = counts.filter((q) => q.queueName === 'EGate').map((q) => q.count).reduce((a, b) => a + b, 0)
+  const deskPax = counts.filter((q) => q.queueName !== 'EGate').map((q) => q.count).reduce((a, b) => a + b, 0)
+  return [egatePax, deskPax]
+}
+
+function* getPaxData(fStart: string, fEnd: string, interval: string, ports: string[]) {
+  const currentResponse: APIResponse = yield call(axios.get, `${ApiClient.passengerTotalsEndpoint}${fStart}/${fEnd}?granularity=${interval}&port-codes=${ports.join()}`)
+  const current = currentResponse.data as TerminalDataPoint[]
+
+  if (ports.includes('LHR')) {
```
ngu04, line 119: Do we need to do accumulate responses for other multiple terminal ports like MAN , BHX  and LGW ?

ngu04 (approved):
Nice one ! Just couple of suggestion if you agreed

react/src/components/regionalpressure/regionalPressureSagas.ts
richbirch: No. LHR is treated as its own region with the terminals being its ports

PR #369, opened 2025-05-22
Drtii 1856 download manager flights export is broken

richbirch:
Implement `FlightsWithSplitsMultiRegionExport` to use for exporting flights from the aggregated flights table


ngu04 (approved):
👍

PR #409, opened 2025-09-22
Drtii 1950 add monitoring around ml jobs

richbirch:
Add support for health checks on the dashboard app itself
Health checks for queue & flight end points
Check json structure in response for expected format


ngu04 (approved):
Nice one !

PR #431, opened 2025-11-25
Drtii 1858 accessibility statement

ngu04:


database_setup/passengers_hourly.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE public.passengers_hourly (
```
richbirch: This table definition is already in drt-lib

database_setup/border_crossing.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE public.border_crossing (
```
richbirch, line 1: Good catch

richbirch (approved):
Nice. We just don't need the passengers_hourly schema here as it's already in drt-lib

PR #446, opened 2025-12-29
Drtii 2013 flight api add queue pax breakdowns

richbirch:
Create API version 1.1
 - add queuePaxCounts array to json for each flight

Add version number to all classes used within a version to avoid ambiguity and allow complete isolation

### DIFF
--- a/src/main/scala/uk/gov/homeoffice/drt/Server.scala
+++ b/src/main/scala/uk/gov/homeoffice/drt/Server.scala
@@ -100,6 +100,9 @@ object Server {
   def dashboardHealthChecks(ports: Iterable[PortCode]): Seq[HealthCheck[_]] = Seq(
     QueueApiV1HealthCheck(SDate.now, ports),
     FlightApiV1HealthCheck(SDate.now, ports),
+
+    QueueApiV1_1HealthCheck(SDate.now, ports),
+    FlightApiV1_1HealthCheck(SDate.now, ports),
   )
 
   private val nonMlPaxPorts = Set("ABZ", "EXT", "HUY", "INV", "LHR", "MME", "NQY", "NWI", "PIK", "SEN")
--- a/src/main/scala/uk/gov/homeoffice/drt/Server.scala
+++ b/src/main/scala/uk/gov/homeoffice/drt/Server.scala
@@ -30,7 +30,9 @@ import uk.gov.homeoffice.drt.ports._
 import uk.gov.homeoffice.drt.ports.config.AirportConfigs
 import uk.gov.homeoffice.drt.routes._
 import uk.gov.homeoffice.drt.routes.api.v1.{AuthApiV1Routes, FlightApiV1Routes, QueueApiV1Routes}
-import uk.gov.homeoffice.drt.services.api.v1.{FlightExport, QueueExport}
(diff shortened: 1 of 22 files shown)


src/main/scala/uk/gov/homeoffice/drt/healthchecks/HealthCheck.scala
```diff
@@ -92,7 +94,27 @@ case class FlightApiV1HealthCheck(now: () => SDateLike, portCodes: Iterable[Port
     "X-Forwarded-Groups" -> (portCodes.map(_.iata).toSeq :+ "api-flight-access").mkString(",")
   )
 
-  override def serialise: String => FlightJsonResponse = _.parseJson.convertTo[FlightJsonResponse]
+  override def serialise: String => FlightJsonResponseV1 = _.parseJson.convertTo[FlightJsonResponseV1]
+}
+
+case class FlightApiV_11HealthCheck(now: () => SDateLike, portCodes: Iterable[PortCode]) extends JsonHealthCheck[FlightJsonResponseV1_1] with FlightApiV1_1JsonFormats {
+  override val priority: IncidentPriority = Priority1
+  override val name: String = "Flight API v1"
```
ngu04: may 1.1 in the name ?  override val name: String = "Flight API v1.1"

src/main/scala/uk/gov/homeoffice/drt/healthchecks/HealthCheck.scala
```diff
@@ -55,7 +57,7 @@ trait PercentageHealthCheck extends HealthCheck[Double] {
     PercentageHealthCheckResponse(priority, name, Failure(new Exception("Failed to parse response")), None)
 }
 
-case class QueueApiV1HealthCheck(now: () => SDateLike, portCodes: Iterable[PortCode]) extends JsonHealthCheck[QueueJsonResponse] with QueueApiV1JsonFormats {
+case class QueueApiV1HealthCheck(now: () => SDateLike, portCodes: Iterable[PortCode]) extends JsonHealthCheck[QueueJsonResponseV1_1] with QueueApiV1JsonFormats {
```
ngu04: may be QueueApiV1HealthCheck needs to be QueueApiV1_1HealthCheck as it serialise V1_1 and extend JsonHealthCheck[QueueJsonResponseV1_1]

src/main/scala/uk/gov/homeoffice/drt/Server.scala
ngu04, line 1: may be should dashboardHealthChecks for V1.1 as well ?

ngu04 (commented):
Nice one ! There is room to reuse and reduce duplication between v1 and v1.1 . May be worth considering it .

src/main/scala/uk/gov/homeoffice/drt/healthchecks/HealthCheck.scala
richbirch: Yeah good catch

src/main/scala/uk/gov/homeoffice/drt/healthchecks/HealthCheck.scala
richbirch: Another good catch - I missed creating a v1.1 class and this was mixed. Fixed now

richbirch:
The duplication is deliberate for 2 reasons:
 - old versions can be easily removed
 - there is no accidental update of multiple versions whilst making a change

PR #452, opened 2026-01-15
Feature/drtii 1919 access to staffing figures change

daryl-boam:
Change 'I require access to staffing figures' checkbox to 'Do you work in your port or command level planning team?' yes no selection buttons.

Added more testing.

### DIFF
--- a/react/src/__test__/integration/AccessRequestForm.spec.tsx
+++ b/react/src/__test__/integration/AccessRequestForm.spec.tsx
@@ -63,7 +63,7 @@ describe('<AccessRequestForm />', () => {
         submitIsDisabled();
     });
 
-    it('enables the submit button when "all ports" is selected and disables it when it is deselected', () => {
+    it('enables the submit button when "all regions" is selected and disables it when it is deselected', () => {
         act(() => {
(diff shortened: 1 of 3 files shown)


react/src/__test__/integration/AccessRequestForm.spec.tsx
```diff
@@ -58,11 +68,12 @@ describe.skip('<AccessRequestForm />', () => {
             render(<AccessRequestForm regions={[{name: 'Heathrow', ports: ['LHR']}]} teamEmail={"test@test.com"}/>);
         });
 
-        fireEvent.click(screen.getByText('All ports'));
+        fireEvent.click(screen.getByText('All regions'));
```
richbirch, line 71: Shall we change the test description so it says 'all regions' rather than 'all ports' now that the label has changed?

richbirch (changes requested):
Nice work. Just a couple of things
* small test description change to match the new label
* the submit button doesn't enable for me when I've ticked all regions and the agreement - we're probably missing a test for this

richbirch (approved):
Lovely. Behaving as expected now :+1:

PR #483, opened 2026-03-25
DRT-II-2140: update to include cirium feed to healthcheck probe

jamessule1:
DRT-II-2140: update to include cirium feed to healthcheck probe


src/main/scala/uk/gov/homeoffice/drt/healthchecks/HealthCheck.scala
```diff
@@ -165,6 +165,27 @@ case class ArrivalUpdatesHealthCheck(minutesBeforeNow: Int, minutesAfterNow: Int
   override def url: String = s"/health-check/received-arrival-updates/${start().toISOString}/${end().toISOString}/$minimumFlights/${updateThreshold.toMinutes}"
 }
 
+case class CiriumFeedHealthCheck(
```
slawomirbedraHO, line 168: Is this class actually called anywhere?
Any chance to have this unit tested?

slawomirbedraHO (changes requested):

